PR #4, opened 2015-10-09
Depends on vue-html-loader explicitly

skyrpex:
Fixes a compilation bug when using node v4.


yyx990803:
Could you also remove the `html-loader` dep? It's replaced by `vue-html-loader`.

skyrpex:
Done!

yyx990803:
Thanks!

PR #8, opened 2015-12-04
Update actions.md

fallenpeace:
Syntax correction.


yyx990803:
Thanks, already fixed in #6

PR #9, opened 2015-12-05
Send the products data in the checkout action

Atinux:
Just to respect the arguments that the checkout option is waiting for:

```js
export const checkout = (products) => (dispatch, state) => {
  const savedCartItems = [...state.cart.added]
  dispatch(types.CHECKOUT_REQUEST)
  shop.buyProducts(
    products,
    () => dispatch(types.CHECKOUT_SUCCESS),
    () => dispatch(types.CHECKOUT_FAILURE, savedCartItems)
  )
}
```


Atinux:
That was blazing fast! :+1:

PR #19, opened 2015-12-11
Missing documentation link.

zigomir:
Documentation is too good to be left out.


yyx990803:
Thanks, I took it down because the API is still unstable. Will put it back up once I settle down with the design a bit more.

PR #53, opened 2016-02-15
Feature/implement getters into store

blake-newman:
Implemented getters into store, mimicking the api of actions.

Getters get passed the store as the first argument and the payload. This is too allow modification payloads for the getters, which allows flexibility on how the getters may be initiated. 

Updated the docs and examples for the implementation details.

### DIFF
--- a/docs/en/SUMMARY.md
+++ b/docs/en/SUMMARY.md
@@ -5,6 +5,7 @@
   - [State](state.md)
   - [Mutations](mutations.md)
   - [Actions](actions.md)
+  - [Getters](getters.md)
 - [Data Flow](data-flow.md)
 - [Application Structure](structure.md)
 - [Middlewares](middlewares.md)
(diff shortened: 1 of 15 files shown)


yyx990803:
Hi Blake,

Thanks a lot for this PR. Really appreciate the effort. One thing I do want to mention is that I happen to be rethinking about Vuex's API lately - I've outlined some planned changes [here](https://github.com/vuejs/vuex/issues/54). Unfortunately, this conflicts with your PR since I am considering changing the way actions are handled.

The good news is that in the proposed new API, `vuex.state` takes exactly the same idea of getters in the form of `state => state.value`. So we don't even need special APIs for getters anymore - it's more like an organization pattern.

blake-newman:
Would it be possible to release this PR, as this will help the current state of Vue. The changes will clearly break this and many other functionality. In the interest of simplifying Vuex for the time being this will be a very minor change in the way things are dealt with. 

It's up to you, but feels worth adding a quick improvement while we discuss and implement a better strategy for Vuex.

yyx990803:
Makes sense :)

PR #63, opened 2016-02-21
Updated counter-hot example.

blake-newman:


yyx990803:
Thanks, with the new API, this is actually not necessary. Changes in actions and getters will bubble up to the components that are importing them, and it will cause those components to hot reload.

blake-newman:
@yyx990803 I tried updating an action and it reloaded the page as it wasn't capturing the file change. Unless this is expected behaviour?

yyx990803:
It needs to be used by a hot-reloadable component.

PR #66, opened 2016-02-28
getters.md - basic example

jbruni:
Prepended a basic example before getting a bit more advanced...

### DIFF
--- a/docs/en/getters.md
+++ b/docs/en/getters.md
@@ -22,7 +22,7 @@ Now `store.getters.total` function can be used in a Vue component similarly to s
 ``` js
 export default {
   computed: {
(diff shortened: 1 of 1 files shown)


docs/en/getters.md
```diff
@@ -1,5 +1,34 @@
...
+  }
+})
+```
+
+Now `store.getters.total` function can be used in a Vue component similarly to state properties:
+
+``` js
+export default {
+  computed: {
+    store.getters.total
```
yyx990803: Should be `total: store.getters.total`

jbruni: Sure

PR #89, opened 2016-03-08
Mechanism to avoid double binding

jbruni:


jbruni:
I was going to open an issue, but decided to open a PR instead.

What is happening to me is: my actions are already binded to the store... when I use them at a Vue instance, I am receiving the `store` twice - as 1<sup>st</sup> and 2<sup>nd</sup> parameters.

Elsewhere, I am already pre-processing the actions' functions, doing `return (...payload) => action(store, ...payload)` (just like in Vuex `0.3.0`). I am using actions heavily outside the VMs...

jbruni:
So, @yyx990803, I just need a way to avoid the "bound" step to be applied again by Vuex, when preparing the actions for the VM. This PR is a suggestion. Maybe you have a better idea of a mechanism to support this.

What I thought was: I can set `action.bounded = true` to the action functions in my code, and then by simply checking it in Vuex to skip the bound, everything else will work as expected. My actions will not be processed again.

I even thought on adding a few more lines to the PR, adding the `bound = true` property to the action functions binded by Vuex itself... who knows if a user may get the `vm` action method and use it in another vm's vuex.actions... (this user would face the same issue I faced, of receiving `store` twice).

This is the *rationale* behind the pull request.

jbruni:
The "few more lines" mentioned would be something like this:

```
  function makeBoundAction (action, store) {
    if (action.bounded) {
      return action
    }
    var vuexBoundAction = function (...args) {
      return action.call(this, store, ...args)
    }
    vuexBoundAction.bounded = true
    return vuexBoundAction
  }
```

yyx990803:
I think this can be done in user land... basically, having two files, `actions.js` and `bound-actions.js` where the latter simply exports bound versions of the raw actions.

jbruni:
And just as a curiosity: it was only when I checked the docs at http://vuejs.github.io/vuex/en/actions.html#calling-actions-in-components that I figured out what was going on. I only did it after spending some time trying to "debug" my app code. I didn't imagine that was expected behaviour. I had to read the doc's *rationale*, and then it made sense. I understand the advantages for the usage inside VMs.

jbruni:
Hmmm... I thought on exporting `actions` (bound) and `vm-actions` (not bound) from the actions module... one for general usage, and other for usage in the components...

jbruni:
I kind of agree with you, but having a way to skip the auto-bound will enable my app-wide code to be so much cleaner, by not having this double set of actions...

jbruni:
Please, give us from "userland" a way (any way) to skip/avoid the "makeBoundAction"... I still wish it a lot. I know no one else gave a +1 by now... but why wait for it when we can provide this immediately?

yyx990803:
I still think this is hardly necessary - why not just

``` js
import store from './store'
import { myAction } from './actions'

myAction(store)
```

blake-newman:
I also agree with @yyx990803 rather than expanding the library to cope with this it would be just as simple to import the store.

This also help with other possibilities,  such as an application that may contain two store instances (reasoning unlikely but still possible). Meaning an action be used by both instances.

jbruni:
I must say that Vuex 2 renders this discussion completely obsolete. I am very happy with the path that Vuex took on its 2.0 version. Enjoying it a lot. Thank you.

PR #95, opened 2016-03-13
Add an option to automatically create actions for mutations

skyronic:
In some apps (for example todomvc itself) there's a use case for quickly
and easily calling a mutation from the controller.

This reduces the effort by automatically creating simple pass-through actions
which automatically camel-case the mutation name and create a function.

Here's my rationale behind this patch: 

http://forum.vuejs.org/topic/3438/vuex-tiny-feedback-suggestion

### DIFF
--- a/examples/todomvc/components/App.vue
+++ b/examples/todomvc/components/App.vue
@@ -66,9 +66,13 @@ export default {
       todos: state => state.todos
     },
     actions: {
-      addTodo,
       toggleAll,
       clearCompleted
+    },
+    auto: function (store) {
+      return {
+        todoActions: store
+      }
     }
   },
   data () {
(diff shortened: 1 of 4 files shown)


skyronic:
This patch isn't complete and there are some known issues. For example, since i can't write a complex object to `methods` it doesn't let me call auto actions from the template. But I'm willing to investigate workarounds for the issues if there's potential to get this features merged.

yyx990803:
Thanks for the PR and feedback! I agree that there could be some way to reduce the amount of steps required to add a single mutation, however I'm not sure if the proposed API is what I'd use. That said, I don't have a very clear idea on what it *should* be like yet. Let's keep the conversation open and see where this goes.

skyronic:
@yyx990803 I agree with you. I was merely exploring how the API would look in production :) 

What about something like this:

```
import { autoActions } from 'vuex/utils'

export default {
   vuex: {
      actions: function (store) {
           return autoActions (store.cart, store.product, { ... extra actions ...});
      }
}
```

PR #113, opened 2016-03-22
Add document explaining how to use websockets

skyronic:
While this is quite straightforward I do believe a lot of people
might have a stumbling block that "how do I trigger an action when no
component is present"


yyx990803:
Since this does not only apply to websockets, I think it'd be better as a sub-section under `actions.md` as "Triggering Actions Outside of Components"?

skyronic:
Ok cool i'll do that.

PR #134, opened 2016-04-04
Better user error.

### LINKED ISSUE #121: Add error handler when action is not a function
If an action is set in the component vuex property as undefined, Vuex will throw an `Uncaught TypeError: Cannot read property 'call' of undefined`. This can happen when importing actions which names have been changed. A better indication of the Action trying to be called and failed would be more helpful.

zigomir:
Proposal on how to fix #121. Feedback appreciated.

### DIFF
--- a/src/override.js
+++ b/src/override.js
@@ -130,12 +130,11 @@ export default function (Vue) {
    */
 
   function makeBoundAction (store, action, key) {
-    return function vuexBoundAction (...args) {
-      if (typeof action === 'function') {
-        return action.call(this, store, ...args)
-      }
-
+    if (typeof action !== 'function') {
       console.warn(`[vuex] Action bound to key 'vuex.actions.${key}' is not a function.`)
+    }
+    return function vuexBoundAction (...args) {
+      return action.call(this, store, ...args)
     }
   }
 
--- a/src/override.js
+++ b/src/override.js
@@ -51,7 +51,7 @@ export default function (Vue) {
       if (actions) {
         options.methods = options.methods || {}
         for (let key in actions) {
-          options.methods[key] = makeBoundAction(this.$store, actions[key])
+          options.methods[key] = makeBoundAction(this.$store, actions[key], key)
         }
       }
     }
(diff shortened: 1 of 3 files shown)


src/override.js
```diff
@@ -126,11 +126,16 @@ export default function (Vue) {
...
    * @param {Store} store
    * @param {Function} action
+   * @param {String} key
    */
 
-  function makeBoundAction (store, action) {
+  function makeBoundAction (store, action, key) {
     return function vuexBoundAction (...args) {
-      return action.call(this, store, ...args)
+      if (Function.prototype.isPrototypeOf(action)) {
```
yyx990803: I think just `typeof action === 'function'` is simpler.

src/override.js
```diff
@@ -126,11 +126,16 @@ export default function (Vue) {
...
+        return action.call(this, store, ...args)
+      }
+
+      throw new Error(`Action bound to key 'vuex.actions.${key}' is not a function.`)
```
yyx990803: Should use `console.warn('[vuex] ...')` to keep things consistent.

zigomir: Any idea how we'd test that case?

yyx990803: You can use a spy on `console.warn` (which means we need Sinon)

zigomir: Cool. I added sinon to spy on console.warn.

src/override.js
```diff
@@ -126,11 +126,16 @@ export default function (Vue) {
    *
    * @param {Store} store
    * @param {Function} action
+   * @param {String} key
    */
 
-  function makeBoundAction (store, action) {
+  function makeBoundAction (store, action, key) {
     return function vuexBoundAction (...args) {
```
yyx990803: actually, we should do the `typeof action` check and warn in `makeBoundAction` before returning the bound action. This way the user gets the warning right at instance creation time instead of when they actually call the action.

zigomir: You're right! I just pushed a fix.

yyx990803:
👍

PR #137, opened 2016-04-07
[Feature] middlewares to ignore 'silent' mutation && FSB support to standards

blake-newman:
- Added support for 'quiet' flag to middlewares
- Fixed FSB standards https://github.com/acdlite/flux-standard-action
- Updated tests
- Updated docs
- Updated API doc to include 'object style mutation'

Note: Modifies FSB support which could mean breaking changes for users.

### DIFF
--- a/docs/en/mutations.md
+++ b/docs/en/mutations.md
@@ -67,6 +67,37 @@ mutations: {
 }
 ```
 
+### Silent Dispatch
+
+> requires >=0.7.0
+
+In some scenarios you may not want the middlewares to record the state change. Multiple dispatches to the store in a short period or polled do not always need to be tracked. In these situations is may be considered appropriate to silent the mutations. 
+
+*Note:* This should be avoided where necessary. Silent mutations break the contract of all state changes being tracked by the devtool. Use sparingly and where absolutely necessary.
+
+Dispatching without hitting middlewares can be achieved with a `silent` flag.
+
+``` js
+/**
+ * Example: Progress action.
+ * Dispatches often for changes that are not necessary to be tracked
+ **/
+export function start(store, options = {}) {
+  let timer = setInterval(() => {
+    store.dispatch({
+      type: INCREMENT,
+      silent: true,
+      payload: {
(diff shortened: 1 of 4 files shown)


blake-newman:
WIP DON'T MERGE, found an issue and need to add tests

docs/en/mutations.md
```diff
@@ -67,6 +67,24 @@ mutations: {
 }
 ```
 
+### Quite Dispatch
```
azamat-sharapov: typo here (and below, in this file)

azamat-sharapov: btw, maybe rename it into something like `silent`? because even English man did a mistake :D

blake-newman: Haha a dyslexic english man, yeah silent is nicer. I'll ammend.

yyx990803:
What are the use cases for silent mutations? I think having silent mutations breaks that contract that all state changes are tracked by the devtool, not sure if that's a good idea unless there's a specifically convincing use case.

blake-newman:
My use case is that I have a progress bar that by the action will dispatch a mutation every so often to update the percent, this floods the devtools and logging. 

```
export function start(store, options = {}) {
  clearInterval(timer);
  store.dispatch(START_PROGRESS_BAR, options.state);

  const cut = 10000 / Math.floor(options.duration || DEFAULT_DURATION);
  timer = setInterval(() => {
    store.dispatch({
      type: INCREASE_PROGRESS_BAR,
      silent: true,
      payload: {
        amount: cut * Math.random(),
      },
    });
    if (store.state.progressBar.percent > 95) {
      store.dispatch(FINISH_PROGRESS_BAR, options.finishState);
    }
  }, options.updateInterval || DEFAULT_UPDATE_INTERVAL);
}
```

For this i really don't care about those mutations, only the directly called actions. 

I'm assured this is not typically best practise to mutate silently. However, in some cases where there is many mutations firing due to an interval, and these mutations generally don't need to be tracked.

PR #146, opened 2016-04-14
Update README.md

jonagoldman:
Add vue-devtools link


ludo237:
That's a fast edit :100:

PR #151, opened 2016-04-20
console.warn when getter is not a function

holmescn:
Someday when I was refactoring my code, I changed some getters name. All things crash with unknown errors. After I digg into the code, and find the reason, I made this PR for you.


yyx990803:
Thanks!

PR #167, opened 2016-04-30
docs/zh-cn/tutorial.md

mondaychen:


mondaychen:
BTW, I had to make a new "vuex flow" image.
If someone else needed the source file to generate the image in the future, it is here: https://drive.google.com/file/d/0B7Q2A8W6dYqDbXlfeWk5YVFWNnM/view?usp=sharing

miljan-aleksic:
The vuex flow images seems have no extension. Can you specify it? I would also suggest creating `arts` folder for this and other vuex art sources.

mondaychen:
@miljan-aleksic This was created using an online tool https://www.draw.io/
Can you see an "open with draw.io Diagrams" button in the google drive page?

miljan-aleksic:
Ah, I see, thanks. If the `arts` folder is accepted we could perhaps save it as .svg

yyx990803:
thanks!

PR #182, opened 2016-05-19
Fix a typo: memoization => memorization

qiansen1386:
:joy:


FadyMak:
Hi @qiansen1386, this is not a typo. Memoization is an optimization technique to reduce the time spent on expensive function calls by caching their results. You can read about [here](https://en.wikipedia.org/wiki/Memoization).

qiansen1386:
Thanks @FadyMak 
I couldn't find it from dictionaries, so I thought that's a typo. Now I have learnt something new! :+1:

Issue #184, opened 2016-05-23
Dispatching object


CthulhuDen:
Documentation says, the whole dispatched object gets sent to the mutation, and so do the release notes for 0.6.2, and says that 

> Note when using the object-style, you should include all arguments as properties on the dispatched object

But in fact only the `payload` field gets passed, and the arguments must be in the payload array, not as some properties. https://github.com/vuejs/vuex/blob/master/src/index.js#L83

dddomin3:
I made a PR to fix the documentation to match the functionality you pointed to here.

CthulhuDen:
On the second thought, it seems even worse: the payload field in the dispatched object is passed as a single argument into mutation (see the code I linked in the first post).

LinusBorg:
> On the second thought, it seems even worse: the payload field in the dispatched object is passed as a single argument into mutation (see the code I linked in the first post).

@CthulhuDen [No. the payload variable will is passed to the mutation using the spread operator](https://github.com/vuejs/vuex/blob/master/src/index.js#L93):

`mutation.forEach(m => m(state, ...payload))`

CthulhuDen:
@LinusBorg But a few line before we see that the payload array will contain only one value (whatever the payload field was)
https://github.com/vuejs/vuex/blob/master/src/index.js#L83

LinusBorg:
Oh, after reading your statement in the comments of the pull request I see what you mean about [[10]] ... odd.

CthulhuDen:
Ok maybe I'm missing something here, please bear with me.

First, you call 

```
store.dispatch({
  type: 'INCREMENT',
  payload: ['firstargument', {second: 'argument'}, 3]
})
```

Then, this (https://github.com/vuejs/vuex/blob/master/src/index.js#L83) happens, so the `payload` is `[['firstargument', {second: 'argument'}, 3]]` (note: this is an array containing array)

So after https://github.com/vuejs/vuex/blob/master/src/index.js#L93 you will have this:

```
mutations: {
  INCREMENT (state, [aString, anObject, aNumber]) {
    state.someString = aString // 'firstargument'
    state.someObject = anObject  // { second: 'argument'}
    state.someNumber = aNumber // 3
  }
}
```
(not that all the payload is in array in second argument)
What am i missing?

LinusBorg:
Yeah I got that right after I finished my comment ... don't really get what's going on there ...

LinusBorg:
Okay, here is my assumption: 
[...]

PR #199, opened 2016-05-31
Fix Object-Style dispatch docs to match actual functionality.

dddomin3:
Issue #184

### DIFF
--- a/docs/en/mutations.md
+++ b/docs/en/mutations.md
@@ -33,13 +33,13 @@ It is also possible to pass along arguments:
 ``` js
 // ...
 mutations: {
-  INCREMENT (state, n) {
-    state.count += n
+  INCREMENT_BY (state, amount) {
+    state.count += amount
   }
 }
 ```
 ``` js
-store.dispatch('INCREMENT', 10)
+store.dispatch('INCREMENT_BY', 10)
 ```
 
 Here `10` will be passed to the mutation handler as the second argument following `state`. Same for any additional arguments. These arguments are called the **payload** for the given mutation.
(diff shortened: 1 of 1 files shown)


docs/en/mutations.md
```diff
@@ -52,17 +52,17 @@ You can also dispatch mutations using objects:
 
 ``` js
 store.dispatch({
-  type: 'INCREMENT',
-  payload: 10
+  type: 'INCREMENT_BY',
+  payload: [ 10 ]
```
CthulhuDen: I'm afraid in case of payload being [10] the mutation argument will be array.

LinusBorg: No it won't as the payload is passed to the mutation using the spread operator

CthulhuDen: It's spread from the array constructed as literal with one element.

stealthybox: @CthulhuDen did you test your statement?
```javascript
console.log( ...[10] )  // prints '10' using node v5.10 and Chrome 51. 
```
@dddomin3 and @LinusBorg are right.
There's no special behavior for a list of one element.

I do think the example could be made clearer with a mutation that accepts multiple args.

CthulhuDen: @stealthybox `console.log(...[[10]])` would be what you should be checking. I did not say that array of 1 element has special handling, but rather that the payload field contents IS the only element of the payloads array later on.

LinusBorg: Where in the code do you get the impression form that it would be a double/nested array?

stealthybox: @CthulhuDen I took a second look -- you're right.

> https://github.com/vuejs/vuex/blob/master/src/index.js#L83
> ```javascript
> payload = [type.payload]
> ```

This doesn't look super intentional.
I don't see a reason to spread an arguments array that will only ever have one element.

stealthybox: If the goal is to encourage users to use an options style object so that payload 'args' are named in the dispatch,
that's achievable using something similar to the previous example. (there's a small typo on `mutations.md` L65)

The strange thing about `index.js` L83 is that it prevents you from using mutations with more than 2 arguments.
An implementation that would allow this would be more appropriate, but it's a breaking change:
```javascript
payload: [10, 'USD']
```
The named parameter example would be changed to: 
```javascript
payload: [{
  amount: 10,
  currency: 'USD'
}]
```

Our team ran into this when we wanted to switch to using Object-Style Dispatches in order to take advantage of the `silent` flag for certain actions.
We think the inability to use previously defined mutations with this style of dispatch is obstructive.

ktsn:
Can we close this PR? Because this behavior seems to be fixed on 1.0.

LinusBorg:
Yeah Evan fixed this some time ago.

PR #201, opened 2016-06-01
support for modules with falsy state

crossjs:
tests:
``` js
  modules: {
    one: {
      state: null
    },
    two: {
      state: undefined
    },
    three: {
      state: false
    },
    four: {
      state: ''
    },
    five: {
      state: 0
    }
  }
})
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -141,7 +141,7 @@ class Store {
 
   _setupModuleState (state, modules) {
     Object.keys(modules).forEach(key => {
-      Vue.set(state, key, modules[key].state || {})
+      Vue.set(state, key, modules[key].hasOwnProperty('state') ? modules[key].state : {})
     })
   }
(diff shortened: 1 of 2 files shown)


LinusBorg:
Hello CrossJS,

and thank you for your contribution. Could you explain a little bit for which usecases this might be useful?

Having a store module  with a single value as state seems odd to me.

crossjs:
Sorry, maybe need more tests, these falsy states seem to be no reactive.

My usecase:
``` js
// progress module
import {
  SET_PROGRESS
} from '../types'

const state = 0

const mutations = {
  [SET_PROGRESS] (state, { payload }) {
    state = payload
  }
}

export default {
  state,
  mutations
}
```

I'll reopen this PR if necessary.

Thanks for your patience.

Issue #215, opened 2016-06-16
Why don't you move the TypeScript definition file to DefinitelyTyped?


ktsn:
I think it would be better to move Vuex's TypeScript definition file to [DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped/).

Because the current Vue's definition file is maintained on DefinitelyTyped. And we can use the definition as dependency.
e.g. We can add `$store` property on the Vue instance definition.

How do you think about this?

kazupon:
I think so too.

ktsn:
@kazupon Thank you!

I'd like to move d.ts file with some improvements if there are no any opposite opinions :)

yyx990803:
It seems the new official type repo is https://github.com/typings/typings

ktsn:
Ah, it's definitely better to register to typings, thank you.

So, should we make a new repository like `typed-vuex` and register it to typings?
I understand we can register the d.ts file in this repository to typings but I think it is noisy to maintain it in this repository.
(It seems to need typings.json, test cases and other stuff of TS for quality)

yyx990803:
Hmm, not sure about the separate repo idea. I'm fine with types living in the same repo if that makes maintenance easier.

yyx990803:
Oh actually, if you mean a separate repo with the type-related files only, that seems ok too.

ktsn:
If you don't care to maintain type-related files in same repository, it's good to do that!
I had worried that the repository becomes mess by type-related files.
But I realize we can separate them to a sub directory and can register it to typings.

I will send a PR to improve the d.ts in this repository and register it later 😄

PR #218, opened 2016-06-19
Update declaration

ktsn:
Related to #215 

I moved all type-related files to `types` directory.
And I added test cases for the d.ts file. You can run the test cases by executing the following command. (It's ok if there are no error is printed)

```sh
$ cd types
$ typings install # install d.ts of Vue.js
$ tsc -p test/tsconfig.json 
```


yyx990803:
Looks great - in the meanwhile, I'm trying to merge `next` branch with `master` for Vuex 1.0 which will support both Vue 1.x and 2.0. Would you mind rebasing after that?

ktsn:
Sure! I will rebase that.

types/index.d.ts
```diff
@@ -0,0 +1,100 @@
+declare namespace Vuex {
+  class Store<S> {
+    constructor(options: StoreOption<S>);
+
+    state: S;
+
+    dispatch(mutationName: string, ...args: any[]): void;
+    dispatch<P>(mutation: MutationObject<P>): void;
+
+    watch(path: string, cb: (value: any) => void, options?: WatchOption): void;
```
yyx990803: I've merged `next` into `master` and this is the only thing that is deprecated. LGTM after removing this.

ktsn:
I've removed it :)

yyx990803:
Thanks! Sorry for the back and forth, but I just pushed some additional changes for 1.0 today. In brief:

- `middlewares` changed to `plugins`. (#233)
- new method: `replaceState(state)`
- new method: `on`, `once`, `off` and `emit`.

Also see (the latest api reference)[https://github.com/vuejs/vuex/blob/master/docs/en/api.md].

ktsn:
OK, I'll add the types for them a little later!

ktsn:
I'm done.

yyx990803:
Great work, thank you very much! Do we need to do anything to publish this to typings?

ktsn:
We need to send PR to [typings/registry](https://github.com/typings/registry) with a json file like [this](https://github.com/typings/registry/blob/master/npm/debug.json).
If you don't mind, I will do that too.

yyx990803:
@ktsn sure!

PR #220, opened 2016-06-20
Allow nested modules

### LINKED ISSUE #185: Nested modules
That would be really helpful for huge platform-ish like apps to have possibility to nest modules, not just have modules of the root store #

ktsn:
This PR fixes #185, allows to define nested modules like the following code.
Properties does not belongs to `state` or `mutations` key are treated as nested modules.

```js
  modules: {
    nested: {
      moduleA: {
        state: {...},
        mutations: {...}
      },
      moduleB: {
        state: {...},
        mutations: {...}
      }
    },
    moduleC: {
      state: {...},
      mutations: {...}
    }
  }
})

console.log(store.state.nested.moduleA)
console.log(store.state.nested.moduleB)
console.log(store.state.moduleC)
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,8 @@
-import { mergeObjects, deepClone, getWatcher } from './util'
+import {
+  mergeObjects, deepClone,
+  isNestedModule, getNestedState,
+  getWatcher
+} from './util'
 import devtoolMiddleware from './middlewares/devtool'
 import override from './override'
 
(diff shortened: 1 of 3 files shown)


yyx990803:
I think this API can be a bit confusing because it relies on implicitness.

What I would do is something like:

``` js
  modules: {
    a: {
      state: {...},
      mutations: {...},
      modules: {
        b: {
          state: {...},
          mutations: {...}
        },
        c: {
          state: {...},
          mutations: {...}
        }
      }
    },
    d: {
      state: {...},
      mutations: {...}
    }
  }
})
```

Basically, each module itself can be declared in the form of `{ state, mutations, modules }`.

ktsn:
Thank you. your solution is reasonable and clearer.
I will fix it :)

ktsn:
It's done.

PR #225, opened 2016-06-24
Update structure.md

whoacowboy:
Removed the word actual, what other type of example would there be?

No big wup, it just sounded funny when I read it, feel free to disregard.


chrisvfritz:
I agree it's an improvement. Thanks!

PR #253, opened 2016-07-19
Allow to register module getters dynamically

### LINKED ISSUE #248: vuex@2.0 getters not registered within dynamic nested modules
Reading `v2.0.0-rc.1` release notes we can see that **getters** should be available inside dynamically registered modules:
> Nested Modules Improvements
Actions, getters and mutations are all supported inside modules.

But when looking in `index.js` source code, into dynamic module registration (`module()` method) I can see that **getters** are not propagated:

`
const {
      state,
      actions,
      mutations,
      modules
    } = module
`
### Desired behaviour
Dynamically registered module getters should be available.

### Current behaviour
http://www.webpackbin.com/E1JVbppL-

I define getters in two ways:
* in `store.js` using `getters: {
    rootState: (state) => state
  }` (when store initializes)
* in `widget.js` using `getters: {
        moduleState : (state) => state
      }` (dynamic module registration, using `store.module(...)`)

As you can see in output `this.$store.getters.moduleState` is empty.

Tested in
* v2.0.0-rc.1
* v2.0.0-rc.3

ktsn:
This patch fixes #248 

The initialization code of internal VM is moved into `initStoreVM` function (renamed from `initStoreState`).
And `initStoreVM` is called on the end of `module` method to register new getters to internal VM.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -20,7 +20,7 @@ class Store {
     this._committing = false
     this._actions = Object.create(null)
     this._mutations = Object.create(null)
-    this._getters = Object.create(null)
+    this._wrappedGetters = Object.create(null)
     this._subscribers = []
     this._pendingActions = []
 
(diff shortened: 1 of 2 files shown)


yyx990803:
Great job!

PR #260, opened 2016-07-28
support array syntax for mapState

chrisvfritz:
Added support for `mapState(['a', 'b', 'c'])` syntax.

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -1,3 +1,5 @@
 machine:
   node:
     version: 5
+  environment:
+    BABEL_ENV: development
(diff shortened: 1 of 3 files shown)


chrisvfritz:
@yyx990803 Finally green! I also added `BABEL_ENV=development` to the Circle CI config so that the tests don't blow up for PRs. I see that you have it set through the website, but environment variables set through the website are assumed to be private, so they're suppressed for PRs.

yyx990803:
@chrisvfritz good catch, thanks!

PR #281, opened 2016-08-15
Avoid destroy watch callback even if resetting inner vm

### LINKED ISSUE #279: vuex@2.0.0-rc.4 - store.watch regression when using it before store.registerModule
##### _Please open examples links one at a time as there are some issues with webpackbin.com when all links are opened in background tabs at once_

When registering modules with `registerModule`, if you add `store.watch` before it, watch callbacks woun't fire (code in `widget.js`).

- [x] Working example using vuex@2.0.0-rc.3 - http://www.webpackbin.com/NJ5S9RFtW
Console logs are shown when you click button and `count` value changes.

- [ ]  Not working example with vuex@2.0.0-rc.4 - http://www.webpackbin.com/Ek--y0FFZ
There are no console logs when you click button and `count` value changes.

But when you change order of `store.watch` and `store.registerModule` callbacks fire as it should.

### Expected result
`store.watch` callbacks fire even when registered before `store.registerModule`

ktsn:
This patch fixes #279.

The watchers for store state are destroyed by `registerModule` because of the reset of the store VM.
This patch adds a new VM `_watcherVM`. It just watches store state and does not be destroyed by `registerModule`.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -23,6 +23,7 @@ class Store {
     this._wrappedGetters = Object.create(null)
     this._runtimeModules = Object.create(null)
     this._subscribers = []
+    this._watcherVM = new Vue()
 
     // bind commit and dispatch to self
     const store = this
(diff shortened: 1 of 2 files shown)


yyx990803:
👍 clever solution

goldfire:
Is this potentially causing a memory leak when used with SSR? I still haven't been able to make a test case as I'm still only finding this issue in production, but in the heap snapshots from production, all of the Vue$2 instances left sitting in memory all point to `_watcherVM in Store`.

PR #325, opened 2016-09-20
Vuex + Vue router example

tobeorla:
I had a hard time making them work together until i found that example.


ktsn:
Hi, @tobeorla!
Could you open a PR to https://github.com/vuejs/awesome-vue instead?
It's the place for this kind of stuff :)

yyx990803:
Hey @tobeorla , that repo hasn't been updated for a while and is likely outdated. We will consider including more info on how to work with the router in the docs.

PR #327, opened 2016-09-22
Support circular structure for logger

### LINKED ISSUE #326: logger.js: Uncaught TypeError: Converting circular structure to JSON
Got an issue with circular structures in leaflet when used inside a store.
There are circular references in leaflet and the store gets JSON.stringified here

Logger.js
Line 20: `var prevState = JSON.parse(JSON.stringify(store.state));`

Here is a store state that causes this:
```
import L from 'leaflet';

const mock = {
  type: 'FeatureCollection',
  features: [
    {
      type: 'Feature',
      properties: {
      },
      geometry: {
        type: 'Polygon',
        coordinates: [
          [
            [
              -0.15483856201171872,
              51.527329038465936,
            ],
            [
              -0.16977310180664062,
              51.51643437722083,
            ],
            [
              -0.15964508056640625,
              51.50094238217541,
            ],
            [
              -0.13149261474609375,
              51.5042549065934,
            ],
            [
              -0.11758804321289061,
              51.518463972439385,
            ],
            [
              -0.13303756713867188,
              51.53106680201548,
            ],
            [
              -0.15483856201171872,
              51.527329038465936,
            ],
          ],
        ],
      },
    },
  ],
};

const layer = L.geoJson(mock);

export default {
  layers: [layer],
};
```

I saw this issue #246 but mine is a different case I think.

Here's the call stack:
image

ktsn:
Fix #326 

The logger plugin throws error if state has circular structure because `JSON.stringify()` cannot handle it.
I added `deepCopy` that handle circular structure and use it in the logger plugin.

I will also open a PR to `next` if this patch is ok.

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "chat": "cd examples/chat && webpack-dev-server --inline --hot --config ../webpack.shared.config.js",
     "build": "node build/build.js",
     "build-examples": "BABEL_ENV=development webpack --config examples/webpack.build-all.config.js",
-    "unit": "BABEL_ENV=development mocha test/unit/test.js --compilers js:babel-core/register",
+    "unit": "BABEL_ENV=development mocha test/unit/*.js --compilers js:babel-core/register",
     "pree2e": "npm run build-examples",
     "e2e": "casperjs test --concise ./test/e2e",
     "test": "eslint src && npm run unit && npm run e2e",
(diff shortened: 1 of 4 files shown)


yyx990803:
Can we add a unit test for this?

ktsn:
Added!
I moved `deepCopy` to util.js to test the function directly. Because it will be complicated to test it from `createLogger`

PR #331, opened 2016-09-24
Fix: arguments for getters missing in API

LinusBorg:
as per `v2.0.0-rc.3`, getters receive 3 arguments: `state, getters, rootState`. However only the first is currently documented. This PR fixes this.


ktsn:
LGTM 👍

PR #345, opened 2016-09-27
fix a small syntax error

romain-web:
`cosnt` into `const`


ktsn:
Thanks for PR, but it's already fixed on #343 :)

Issue #335, opened 2016-09-25
Vuex 2.0 module's actions and getters namespacing


jxlarrea:
The documentation is quite vague regarding namespacing actions and getters from a module. Can someone please elaborate on how to achieve this?

Namek:
I guess it's something like this:
```
export const UPDATE_PROPERTY = '*/UPDATE_PROPERTY'

export const CREATE_DAY_DATA = 'day/CREATE_DAY_DATA'
export const LOAD_DAY_DATA = 'day/LOAD_DAY_DATA'

export const FILTER_PRODUCTS = 'products/FILTER_PRODUCTS'
export const SELECT_PRODUCT = 'products/SELECT_PRODUCT'
```

where `/` splits between module name and mutation/action type name.

jxlarrea:
@Namek honestly that convention doesn't seem too elegant. It would be good to have in the documentation some best practices when it comes to namespacing.

Namek:
Well, then I don't know what kind of "namespacing" the docs refer. I got this from some older examples.

ktsn:
Namespacing in the doc means adding prefix or suffix to getters, actions and mutations names to avoid name conflict.

jxlarrea:
@ktsn Do you have some sample code with prefixed actions, getters, etc.?

ktsn:
Like this? https://jsfiddle.net/bjahdbx5/1/

jxlarrea:
Thank you for the code. That is exactly the result I'm trying to achieve. But it does seem that it would get extremely verbose adding this code to all actions, getters and mutations per module. Is there any way to automate the generation of the prefix either in the module.js file?

ktsn:
Do you mean that you do not want to write `namespace` function each definition?
If so, you can just rewrite the helper function.
https://jsfiddle.net/bjahdbx5/2/

Vuex does not concern about how to define getters, actions and mutations names since it is userland issue.
We should create own helpers as our needs if we suffer to define the constants.

jxlarrea:
Thanks for the help @ktsn. Your code helped me to create a helper utility to namespace getters, actions and mutations in Vuex 2.0 modules. This functionality really should be included in Vuex itself. Large application need this, and leaving this for the user to deal with only adds a huge amount of unnecessary boilerplate code.

ktsn:
I think it should not be included into Vuex because it can be solved by external library and using constants is optional.
But anyway, the description of namespace in docs can be improved as you said.
[...]

PR #348, opened 2016-09-27
[doc] Add more concrete description for namespace

ktsn:
The current namespace description is vague as pointed out on #335.
I've added that "namespace" means adding prefix or suffix to names of getters, actions or mutations, and added an example code of it.


LinusBorg:
LGTM

YamiOdymel:
Sorry but how do we use namespace in the components?

PR #357, opened 2016-10-02
change "INCREMENT" to "increment"

zengxiaotao:


kobione:
Good job catching the error, zengxiaotao.

I suggest changing all to "INCREMENT" as the convention for mutations is uppercase.  It is particularly important to do so in these docs, as  the lack of consistent use of uppercase, is divergent from all the other documentation. 

Newer developers may be learning to use Vue2/Vuex2, and the lack of uppercase adds confusion. (Is this not the point of the convention: to easily pick out the mutation at a quick glance?) 

If we are going to have a convention in the community, best-practices should be followed in the documentation.  (I would attempt to fix myself, but do not want to cause errors).

zengxiaotao:
@kobione yeah,  change all mutation types to uppercase is better

PR #368, opened 2016-10-06
Update modules.md

blake-newman:


miljan-aleksic:
I was about to PR about this as well. The second argument is actually the payload, if any.

PR #375, opened 2016-10-08
Fixed a missing semi-colon in the code example

webnoob:
The code example provided didn't contain a semi-colon where required.


ktsn:
Thanks for PR.
However, colons must not be included in that place because the `mapGetters` is not property name but a function.

webnoob:
Whoops! Sorry.

PR #397, opened 2016-10-14
changed router to Vuex, as this isn't the router

dsikes:
Im not sure if this is a typo or not. This wording makes more sense to me as I am reading through your docs though. I use React a good bit in other projects.. Still getting used to Vue. Great work.


ktsn:
I think it's typo indeed. Thanks.

PR #401, opened 2016-10-16
Fixed symlink problem on Windows & translate docs to zh-cn

jingsam:


jingsam:
Symlinks are not working on Windows

ktsn:
Why would you want the docs to work on Windows?
It seems that there are no reason to do that because we can read it on the web with any OS.
And it will increase maintenance cost without symlink...

In addition, we cannot merge the PR that includes two or more large fixes, please separate them.

jingsam:
I am working on translating this docs to a Chinese version and almost fininshed. I am a Windows user. Symlinks can not compile by gitbook on Windows. What'more, the contents of book.json in en and zh-cn version are different. So you need seperate them rather than using symlinks.

jingsam:
Does that make sense? @ktsn If it does, I have send another PR #406

jingsam:
@ktsn I have finished the Chinese version docs. Maybe you can  invite someone to review it, thanks

yyx990803:
Merged as eaf53da - thank you!

Issue #380, opened 2016-10-09
New module option `namespace`


ktsn:
> Last update: 2016-10-23

This proposal is come up from the discussion in #359.
### Summary of proposal

Modules will have `namespace` option that expects string value. All getters, actions and mutations types in the module are prefixed by the given value.

Namespace is inherited to child modules implicitly and we can also nest namespace by declaring `namespace` option on the child modules.

Also, this option can be more useful with `subStore` option that is another new module option proposal #381.
### Why is this worth adding?

Currently Vuex does not auto-namespace getters, actions and mutations even though they are defined in some nested modules, the namespacing is handled by developers. This design is reasonable because there are some cases that the auto-namespacing will suffer developers as discussed in #236 .

However, I think it would be useful that we have optional auto-namespacing because we probably want to prefix all getters, actions and mutations type in the same module in most cases.
### Examples

``` js
export default {
  namespace: 'user/',

  // module assets
  state: { ... }, // module state will not be changed by prefix option
  getters: {
    isAdmin () { ... } // -> getters['user/isAdmin']
  },
  actions: {
    login () { ... } // -> dispatch('user/login')
  },
  mutations: {
    setName () { ... } // -> commit('user/setName')
  },

  // nested modules
  modules: {
    // inherit the prefix from parent module
    nestedA: {
      state: { ... },
      getters: {
        nestedGetterA () { ... } // -> getters['user/nestedGetterA']
      }
    },

    // nest the namespace
    nestedB: {
      namespace: 'nestedB/',

      state: { ... },
      getters: {
        nestedGetterB () { ... } // -> getters['user/nestedB/nestedGetterB']
      }
    }
  }
}
```

yyx990803:
I like this.

One thing about nested modules with prefix - should it become `nestedB/xxx` or `user/nestedB/xxx`? I feel that multiple nesting makes more sense.

psi-4ward:
definitely `user/nestedB/xxx` as this one is expected when nestedB is a module from user

ktsn:
Because I thought if we want to register a root level mutation, we can clear parent prefix by overwriting.
But I've noticed we had better to define such mutation on root and proxy to the nested mutation. I update about the nested prefix :)

LinusBorg:
> One thing about nested modules with prefix - should it become nestedB/xxx or user/nestedB/xxx? I feel that multiple nesting makes more sense.

But the is would mean that 
[...]

Issue #381, opened 2016-10-09
New module option `subStore`


ktsn:
This proposal is come up from the discussion in #359.
### Summary of proposal

Modules will have `subStore` option that expects a sub store name. If the option is specified, a sub store will be created under `store.subStores`.

The sub store has `state`, `getters`, `dispatch` and `commit` which are same interface as `Vuex.Store` but state will be the local state of corresponding module.

If `prefix` option (another proposal #380) is specified, sub store will prefix all getters, actions and mutations types implicitly. For example, if we set `prefix: 'user/'` on sub store module, `subStore.dispatch('login')` will be transformed to `rootStore.dispatch('user/login')`.
### Why is this worth adding?

Currently, we need to collect state, getters, actions and mutations from global namespace when we use them on components even though we separate the application state and logics to each module.

It would be good to provide the sub store because we can load all assets in a module at once and use them without prefix.
### Examples

Module definition:

``` js
const userModule = {
  subStore: 'user',
  prefix: 'userModule/'

  // module assets
  state: { ... },
  getters: {
    isAdmin () { ... }
  },
  actions: {
    login () { ... }
  },
  mutations: {
    setName () { ... },
  }
}
```

Usage of sub store:

``` js
const store = new Vuex.Store({
  modules: {
    user: userModule
  }
})

const user = store.subStores.user
user.state.id // -> store.state.user.id
user.getters.isAdmin // -> store.getters['userModule/isAdmin']
user.dispatch('login') // -> store.dispatch('userModule/login')
user.commit('setName') // -> store.commit('userModule/setName')
```

Component binding:

``` js
computed: {
  // 2nd argument will be sub store name
  ...mapGetters({
    admin: 'isAdmin' // -> admin: 'userModule/isAdmin'
  }, 'user')
}
```

yyx990803:
I think this has overlaps with modules and makes things more complex than they should be. I'm not convinced about the benefits vs. the extra complexity.

psi-4ward:
So I currently try to get my head around the vuex stuff (hard btw).
As I understood its the single-source-of-truth so splitting it up results back in multie-sources-of-truth?

ktsn:
@yyx990803 
I think this feature is not overlap with modules because modules let us divide the concern on *definition time* but they are flatten when we use them. So we need to write prefix for all execution of getters, actions and mutation. This is what the sub store would solves.

@psi-4ward 
[...]

PR #420, opened 2016-10-24
Implement module `namespace` option

ktsn:
This PR is an implementation of #380 with update for docs and typings.
### Features
- Modules can have `namespace` option that expects `string` value.
- This option allows us to add prefix for all getters, actions and mutations in the same module, and also can inherit or nest a parent prefix (https://github.com/vuejs/vuex/issues/380#issue-181871602).
- `getters`, `dispatch` and `commit` in a namespaced module's action context will be localized. i.e. implicitly add module's namespace on their execution (https://github.com/vuejs/vuex/issues/380#issuecomment-252629629).
- `dispatch` and `commit` can receive `root` option on their last argument. If `root` option is specified, they dispatch/commit on global level (See example code in https://github.com/vuejs/vuex/issues/380#issuecomment-252629629).
- Getter functions in a namespaced module will receive localized `getters` like action context.
- Getter function will receive `rootGetters` on 4th argument.
### Implementation Details

I've add `ModuleCollection` and `Module` class to manage Vuex modules because it would be complicated to handle namespacing by plain objects. For example, it is difficult to detect runtime module's namespace on `registerModule` or `resetStore` because `_runtimeModules` is separated from the entire module structure.

In `ModuleCollection`, each `Module` is stored as tree structure that is same as the structure of modules option.

To localize `getters`, `dispatch` and `commit`, I added a proxy object and functions. The local `dispatch` and `commit` just add namespace string to action/mutation type. The local `getters` implementation is a bit tricky. To create the local `getters`, I iterate `store.getters` keys and take matched keys to the proxy object.

### DIFF
--- a/docs/en/modules.md
+++ b/docs/en/modules.md
@@ -156,6 +156,22 @@ export default {
 }
 ```
 
+#### Caveat for Plugin Developers
+
+You may care about unpredictable namespacing for your modules when you create a [plugin](plugins.md) that provides the modules and let users add them to a Vuex store. Your modules will be also namespaced if the plugin users add your modules under a namespaced module. To adapt this situation, you may need to receive a namespace value via your plugin option:
+
+``` js
+// get namespace value via plugin option
+// and returns Vuex plugin function
+export function createPlugin (options = {}) {
+  return function (store) {
+    // add namespace to plugin module's types
+    const namespace = options.namespace || ''
+    store.dispatch(namespace + 'pluginAction')
+  }
+}
+```
+
 ### Dynamic Module Registration
 
 You can register a module **after** the store has been created with the `store.registerModule` method:
--- a/docs/en/modules.md
+++ b/docs/en/modules.md
@@ -81,42 +81,77 @@ const moduleA = {
 
 ### Namespacing
 
-Note that actions, mutations and getters inside modules are still registered under the **global namespace** - this allows multiple modules to react to the same mutation/action type. You can namespace the module assets yourself to avoid name clashing by prefixing or suffixing their names. And you probably should if you are writing a reusable Vuex module that will be used in unknown environments. For example, we want to create a `todos` module:
+Note that actions, mutations and getters inside modules are still registered under the **global namespace** - this allows multiple modules to react to the same mutation/action type. You probably should namespace your Vuex module if you are writing a reusable one that will be used in unknown environments. To support namespacing for avoiding name clashing, Vuex provides `namespace` option. If you specify string value to `namespace` option, module assets types are prefixed by the given value:
 
 ``` js
-// types.js
+export default {
+  namespace: 'account/',
 
-// define names of getters, actions and mutations as constants
-// and they are prefixed by the module name `todos`
-export const DONE_COUNT = 'todos/DONE_COUNT'
-export const FETCH_ALL = 'todos/FETCH_ALL'
-export const TOGGLE_DONE = 'todos/TOGGLE_DONE'
+  // module assets
+  state: { ... }, // module state will not be changed by prefix option
+  getters: {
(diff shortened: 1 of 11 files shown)


Kaijun:
Does it mean that not only states but also getters/actions will also be modularized?

Current version of Vuex only supports modularization of states.

ktsn:
@Kaijun Yes and no. This option allows us to just add prefix to types of getters, actions and mutations in the same module. They are still flattened on registration.

Kaijun:
@ktsn hmmm... one more question, so if two actions are sharing same name but under different modules, so the first action will still be overridden by the second action?

ktsn:
@Kaijun No, both actions are triggered by `dispatch` as same as the current behavior.

Kaijun:
will it be merged?

YamiOdymel:
I would like to see this get merged, 

otherwise I can only use the `namespace.js` helper utility which is mentioned in #335.

kiaking:
One advantage I find in `namespace.js` helper at #335 compared to this proposal is that it lets you allow committing or dispatching by constant variables. For example, you can do this;

```js
import * as types from './types'

this.$store.dispatch(types.someAction)
```

It seems to me that with this PR we need to use string for the argument. Maybe string is the way to go, but do you have any thoughts on this?

ktsn:
@kiaking I think we can use constant by #381

kiaking:
@ktsn Oh yeah I see. Thank you for the answer. Can't wait to see this get merged (with #381).

skyrpex:
Maybe this is a little bit offtopic but... would make it sense to be able to use a Symbol as a mutation key?

vinicius73:
any news about this?

LinusBorg:
@ktsn We should try and get this merged if you think it'ready.

ktsn:
@LinusBorg Yeah, it's ready. 

@yyx990803 Can you review this?

yyx990803:
@ktsn sorry for taking such a long time to review this - it looks great!

yyx990803:
Although this is already merged - there's one thing I'd like to revisit: is it necessary for the `namespace` option to be a string? Since each module is already uniquely identified by its state path, it seems redundant to have to declare the string. Instead, we can just make the option `namespaced` and expect a boolean value:

``` js
modules: {
  foo: {
    namespaced: true, // infers the `foo/` namespace automatically
    state: { ... }
  }
}
```

In addition, the use of a string namespace option indicates there is the possibility that the namespace of a module to be different from its state path - which in my opinion should be avoided.

Unifying the namespace and the state path also makes it possible to enhance the `mapXXX` helpers with a unified API, e.g.:

``` js
computed: {
  ...mapState('foo/', {
    bar: state => state.bar // state is the state of the `foo/` module instead of root
  }
},
methods: {
  ...mapActions('foo/', {
    // ...
  })
}
```

ktsn:
Sounds good to me. 
I proposed namespace option as string because I want to reduce the nest of namespace for deeper module to use it simply in components. 
But it seems to be resolve the unification and new mapXXX helpers´ feature. (and I can drop #381)

Also we probably simplify the namespace implementation :)

PengMQ:
hey, can we have the 'namespace' option set for module, and then in component, when I do mapState or mapActions, there could be a param as 'namespace' to them. By doing so, I can map the store and the component, so I will not get the 'duplicate xx key error' without add any prefix to my getters, my actions...

LinusBorg:
Please read the documentation for mapState etc., this is already possible.

PengMQ:
Do you mean the mapping of  'namespace option of store' and the 'namespace param of mapState...' is already possible? or just the 'namespace param of mapState'?

LinusBorg:
The latter. How would you imagine this "mapping" should work?

PR #448, opened 2016-11-08
[docs] Align GitBook theming to main Vue.js site

pearofducks:
One of the strengths Vue highlights is that these extra plugins are maintained by the core team. With this change, the documentation feels a lot more like this is true. :)

Note the lack of chapter numbers is not my doing, that seems to possibly be a change from GitBook 2 to 3 - but I do think it's better without them.

> edit// Also, I'd recommend bumping the GitBook version in `book.json` if possible to `>3.0.0`, I couldn't get the current version to run locally. Let me know if this an OK change and I can add it here.

before

after


ktsn:
Looks good to me.
@yyx990803 What do you think?

kazupon:
awesome. 💯 
if this theme is provided as the gitbook plugin, we can use it into the other vue docs.
we should be made the gitbook theme plugin.

yyx990803:
This is great! And +1 to what @kazupon said

pearofducks:
Plugin made! This applies the css globally - except the language-selection screen (a limitation of GitBook).

yyx990803:
@pearofducks thank you so much! Would appreciate PRs for vue-router and vue-loader as well :)

PR #457, opened 2016-11-14
[Patch] Remove silent option from mutations

blake-newman:
- Now Dev tools supports filtering silent option is no longer required
- Updated unit and type tests


ktsn:
Maybe we should deprecate the option and provide warning for users at first?

yyx990803:
Yeah, this would technically be a breaking change. Let's mark it deprecated for now.

blake-newman:
Good idea, would you prefer to depreciate that feature (silent flag) or the options object. 

As there is no need for the object at all in with depreciation, however technically we are depreciating just the silent flag.

ktsn:
I think it would prefer to deprecate silent flag only, because I introduce new option in #380  😉

blake-newman:
@ktsn I will modify to keep option parameter available and depreciate the silent flag.

#380 is fascinating!

src/index.js
```diff
@@ -78,8 +78,16 @@ class Store {
         handler(payload)
       })
     })
-    if (!options || !options.silent) {
-      this._subscribers.forEach(sub => sub(mutation, this.state))
+    this._subscribers.forEach(sub => sub(mutation, this.state))
```
ktsn: I think we should not only provide deprecation message but also retain the silent flag's behavior until next major release.

src/index.js
```diff
@@ -78,8 +78,16 @@ class Store {
...
+
+    if (
+      process.env.NODE_ENV !== 'production' &&
```
ktsn: As vuex currently does not resolve `NODE_ENV` check on the build time, we should not introduce it?
This line would cause runtime error if we use Vuex from `<script>` tag.
Or can we update build configuration for it?

blake-newman:
@ktsn to be honest this feature is rarely used, so probably no need to check env variable. 

Will modify to still retain existing silent functionality if @yyx990803 agrees

PR #482, opened 2016-11-25
gitbooks not showing link for "Core concepts"

thewhitetulip:
While reading the online book there is no link for next chapter after the Getting started one because Core concepts has no file attached to it.


ktsn:
It seems like there is still not exist the link though.

thewhitetulip:
@ktsn I am really sorry I didn't understand what your comment meant.

ktsn:
> there is no link for next chapter after the Getting started

I mean that this situation persists when I check the docs with your patch in local.

thewhitetulip:
Did the gitbook's build fail? I can't really find the root cause of this not working without seeing the output of gitbooks. Do you have gitbooks installed on local?

ktsn:
The build succeeded but there is still no link for next page in the Getting Started page. Or you just intended to add the link at "Core Concept" in ToC?

thewhitetulip:
Well, I am not sure what I did :-)

So let me pinpoint the problem so you can fix it. 

When I am reading the gitbooks book, I don't have a (next) button for every page except http://vuex.vuejs.org/en/getting-started.html has a next button. This is because the next section has no link, thus no next.

ktsn:
OK, I found the cause. We should update `SUMMARY.md` to available the next button in Getting Started.
But if we do so, the next button in State chapter is not working. I guess it is because the link for `state.md` is duplicated on ToC.
Do you have any ideas to solve this? I'm actually not familiar with gitbook :)

thewhitetulip:
Yes, you are right. I have used gitbooks earlier once, https://github.com/thewhitetulip/web-dev-golang-anti-textbook/blob/master/SUMMARY.md

What I had done to mitigate this issue is add a new file for the one not showing the link, like a landing page. Can we do the same? Or what we can do is, shift the `state` one level up and link to it in Core Concepts itself.

Do let me know what you wish to do. We can close this issue and start a new one.

ktsn:
Thanks for suggesting your solutions.

> add a new file for the one not showing the link, like a landing page

I'm not sure if I understand you correctly but do you mean adding `core-concepts.md` that including links for child chapters (State, Getters, etc.)?
If so, I think it would be the best idea. 🙂

thewhitetulip:
Yes, we can add a new core-concepts.md which will just say something like , " In this section we will study the core components". with a link to the child chapters :-)

ktsn:
Great! Can we add it on this PR?

thewhitetulip:
Done! @ktsn Please check.

docs/en/core-concepts.md
```diff
@@ -0,0 +1,10 @@
+# Core concepts
+
+We will be learning the core concepts in Vuex in this chapters. They are
+- State
+- Getters
+- Mutations
+- Actions
+- Modules
```
ktsn: It would be great if these items have the link to the corresponding chapter :slightly_smiling_face:

docs/en/README.md
```diff
@@ -7,7 +7,7 @@
 - [Installation](installation.md)
 - [What is Vuex?](intro.md)
 - [Getting Started](getting-started.md)
-- Core Concepts
+- [Core Concepts](state.md)
```
ktsn: Can we also update this link to `core-concepts.md`?

docs/en/core-concepts.md
```diff
@@ -0,0 +1,10 @@
+# Core concepts
```
ktsn: It's a trivial thing but camelizing all words is better like other chapter title.
`Core Concepts`

docs/en/README.md
thewhitetulip: Oh, sorry. I had intended it to have core-concepts.md link only. Somehow missed it.

thewhitetulip:
:+1: Done

ktsn:
Thanks! LGTM :slightly_smiling_face: 
/ping @yyx990803

thewhitetulip:
I had to search for what LGTM meant. :smile:

michalsnik:
Hey, is the a reason it's not merged yet?

ktsn:
Sorry for leaving this for a long time. This look great, thank you!

thewhitetulip:
Finally!

PR #492, opened 2016-11-29
Update store.watch() typing

plievone:
watch callback gives oldValue as the second parameter.
It may also be undefined in case `{immediate: true}` is used, but typing it optional could be inconvenient for most common uses.


ktsn:
I think it would be a reasonable. Thanks.

PR #495, opened 2016-11-30
Added example of using namespaced types

wolthers:
This PR adds an example of how you would go about implementing enumerable types for your vuex actions, getters and mutations with namespace prefixes so they don't collide in the global vuex namespace.

### DIFF
--- a/examples/namespaced-types/index.html
+++ b/examples/namespaced-types/index.html
@@ -3,7 +3,6 @@
   <head>
     <meta charset="utf-8">
     <title>vuex namespaced types example</title>
-    <link href="http://fonts.googleapis.com/css?family=Muli" rel="stylesheet" type="text/css">
   </head>
   <body>
     <div id="app"></div>
--- a/examples/namespaced-types/index.html
+++ b/examples/namespaced-types/index.html
@@ -1,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>vuex namespaced types example</title>
+    <link href="http://fonts.googleapis.com/css?family=Muli" rel="stylesheet" type="text/css">
+  </head>
+  <body>
+    <div id="app"></div>
+    <script src="/__build__/shared.js"></script>
+    <script src="/__build__/namespaced-types.js"></script>
+  </body>
+</html>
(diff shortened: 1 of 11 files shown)


examples/namespaced-types/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8">
+    <title>vuex namespaced types example</title>
+    <link href="http://fonts.googleapis.com/css?family=Muli" rel="stylesheet" type="text/css">
```
ktsn: Why do you load this font?
Seems like it isn't used.

wolthers: Hey @ktsn - I just copied another example, but I can see it isn't used. Will remove it.

wolthers:
Cool. My motivation for doing this was the lack of examples in the namespace part of the docs of how you would go about mapping a getter/action to a component (e.g. "add/ADD_TO_CART"). If you use the array form - `...mapGetters(['add/ADD_TO_CART'])` - it fails, so you have to alias it using the object form: `...mapGetters({addToCart: 'add/ADD_TO_CART'})`. 

So, updating the docs would probably be a good idea also. I don't know all those languages though, so I didn't do it myself.

ktsn:
You can update the English (or your familiar language) docs at first. So someone can translate it to other languages.

wolthers:
It was easy enough to update for all locales, so I did it.

yyx990803:
Hey @wolthers , thanks for the PR, although I'm wondering if this is still relevant after merging #420 ? @ktsn what do you think?

Once 2.1 ships with the namespace option, I think we should recommend everyone using that instead of manually namespacing modules.

wolthers:
@yyx990803: i still think the docs need to mention how to map namespaced getters and actions to a component as this was the point of this PR. How namespacing is implemented is minor for me, as long as it's well documented.

ktsn:
I thought the manual namespacing can be option because the initial idea of #420 does not solve the pain of using getters/actions/mutations in components. But now I think we can probably migrate to namespace option by [the namespace unification and new mapXXX feature](https://github.com/vuejs/vuex/pull/420#issuecomment-266312694).

LinusBorg:
@wolthers @ktsn this PR is no longer required now that #420 was merged and released with [v2.1.0](https://github.com/vuejs/vuex/releases/tag/v2.1.0), isn't it?

ktsn:
Yes, we should not do manually namespacing since v2.1.0. It would be better to use the `namespaced` option and pass a namespace string to component binding helpers.

@wolthers I'm sorry for closing this PR after I approved this. Thank you for your PR though!

PR #503, opened 2016-12-04
use README content importing with include tag

kazupon:
reference gitbook toolchain documentation 
http://toolchain.gitbook.com/templating/conrefs.html

I think `include` tag is DRY.


ktsn:
@kazupon Thanks! This looks perfect 👍

PR #496, opened 2016-11-30
Add yarn command for install Vuex [RU]

eskrano:

PR #505, opened 2016-12-05
Add yarn-way installation docs Vuex [all langs]

gbezyuk:
See #496 for original request. (Surprisingly, can't create a PR for that PR's patch-1 branch, so I do it right here)


ktsn:
@gbezyuk Thanks 🙂

Issue #460, opened 2016-11-16
Access store.getters in the watched getter function created for store.watch?


dyreby:
Currently store.watch takes a getter function that receives store.state as its only argument.  Would it be possible to also pass in the store.getters object as an argument to this newly created watched function?

The following would then be possible:

~~~~
store.watch((state, getters) => {
  if (state.firstChoice) return getters.someComplicatedGetter
  else return getters.someOtherGetter
}, newValue => {
  console.log('callback:', newValue)
})
~~~~

The reasoning is that one may want to watch for changes in a number of complicated getters defined and used elsewhere.

Thanks much for considering this idea.

ktsn:
I think it's good idea. 👍

LinusBorg:
added a PR for this.

ktsn:
It's already merged :)

PR #515, opened 2016-12-10
Add store getters as second argument to store.watch() computed fn (solve #460)

LinusBorg:
* the first parameter of store.watch(), which is a function, now receives the store.getters object as its second argument.
* added unit test

Fullfills request #460

### DIFF
--- a/test/unit/test.js
+++ b/test/unit/test.js
@@ -5,7 +5,7 @@ import Vuex, { mapState, mapMutations, mapGetters, mapActions } from '../../dist
 Vue.use(Vuex)
 
 const TEST = 'TEST'
-const TEST2 = 'TEST2'
+// const TEST2 = 'TEST2'
 
 describe('Vuex', () => {
   it('committing mutations', () => {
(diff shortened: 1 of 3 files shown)


test/unit/test.js
```diff
@@ -5,7 +5,7 @@ import Vuex, { mapState, mapMutations, mapGetters, mapActions } from '../../dist
 Vue.use(Vuex)
 
 const TEST = 'TEST'
-const TEST2 = 'TEST2'
+// const TEST2 = 'TEST2'
```
LinusBorg: eslint complained that this constant was never used anywhere in the tests so i commented it out. Should I remove it, or is it needed for something? Tests all worked without it.

ktsn: I think we can remove it.

LinusBorg:
@ktsn removed it.

yyx990803:
@LinusBorg looks like the dist file is still included in the PR

LinusBorg:
Oh right, sorry

LinusBorg:
@yyx990803 I reversed the changes to /dist.

PR #534, opened 2016-12-20
Update modules.md

davidmsvetlecic:


ktsn:
Actually, this should be `getters['account/profile']` because the `myPage` module does not have `namespaced` option.

kazupon:
oh, sorry

PR #540, opened 2016-12-23
Save dev is 90% of the time more usefull than nothing

CoolGoose:


fnlctrl:
Should be `--save` since vuex isn't a `devDependency`.

docs/en/installation.md
```diff
@@ -18,7 +18,7 @@ Include `vuex` after Vue and it will install itself automatically:
 ### NPM
 
 ``` bash
-npm install vuex
+npm install vuex --save-dev
```
LinusBorg: as mentioned by @ktsn, this should be `--save`

CoolGoose:
@LinusBorg --save it is ;)

PR #553, opened 2016-12-30
Housekeeping eslint for v2.0

ktsn:
Same change as #551 but for v2.0


kazupon:
In circeci, occurred E2E test errors.

ktsn:
It's version mismatch between Vue and vue-template-compiler, I've rebuilt CI.

PR #555, opened 2016-12-31
accept a function for the initial state

yanick:
More of a RFC than a real PR, as it needs tests and documentation, but a naive stab at dealing with cases where we'd like to have a fresh state each time we create a new store.

For example:

```
import Vuex from 'vuex';
import Vue from 'vue';

Vue.use(Vuex);

let baz = {
    state: { quux: 21 },
    mutations: { 'inner': state => state.quux++ },
};

let new_store = () => new Vuex.Store({
    state: () => { foo: 12 },
    mutations: {
        'bar': state => state.foo++
    },
    modules: { baz },
});

let store = new_store();

store.commit('inner');
store.commit('bar');

console.log( JSON.stringify( store.state ) );
// { foo: 13, quux: 22 }

store = new_store();

store.commit('inner');
store.commit('bar');

console.log( JSON.stringify( store.state ) );
// OOPS! { foo: 13, quux: 23 }
```

To get around that, I allow for the initial state to be generated by a given function, like it is done in Redux.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -11,6 +11,8 @@ class Store {
     assert(Vue, `must call Vue.use(Vuex) before creating a store instance.`)
     assert(typeof Promise !== 'undefined', `vuex requires a Promise polyfill in this browser.`)
 
+    if ( typeof options.state === 'function' ) options.state = options.state();
+
     const {
       state = {},
       plugins = [],
@@ -232,7 +234,11 @@ function installModule (store, rootState, path, module, hot) {
     const parentState = getNestedState(rootState, path.slice(0, -1))
     const moduleName = path[path.length - 1]
     store._withCommit(() => {
-      Vue.set(parentState, moduleName, module.state)
+      Vue.set(
+        parentState,
+        moduleName, 
+        ( typeof module.state === 'function' ) ? module.state() : module.state
+      )
     })
   }
 


ktsn:
Thanks for your PR.
Why do you create multiple stores? As Vuex is designed with single source of truth pattern, you should not do that.
I think we can take other solutions for your original issue.

yanick:
> Why do you create multiple stores? As Vuex is designed with single source of truth pattern, you should not do that.

I have two scenarios where many stores are created. 

1. The first one is simply for testing, where I do create the store, dispatch a few things, then discard and create a new one to test other stuff. The example in the PR description is a boiled down version of those tests.

2. I'm working with app that, on the server-side, will arbitrate many turn-based games. Each game is an individual and independent store.

Mind you,  it's true that I could get around the problem by changing the example to do something like

```
let baz = () => ({
    state: { quux: 21 },
    mutations: { 'inner': state => state.quux++ },
});

let new_store = () => new Vuex.Store({
    state: { foo: 12 },
    mutations: {
        'bar': state => state.foo++
    },
    modules: { baz: baz() },
});
```

Which is not a huge difference (basically, always make sure that the modules used by the new stores are fresh objects with fresh states).  I just though that having allowing the state to be generated just like the `data` field for Vue components could be nice. And it gave me an excuse to peek under the hood of VueX. :-)

ktsn:
It makes sense for testing purpose. However, I think just wrapping by a function is a better solution because we can set various state, actions, getters and mutations for each test case.

> I'm working with app that, on the server-side, will arbitrate many turn-based games. Each game is an individual and independent store.

I think we should use only store even if in such case.

1. Generally speaking, there would be a demand to cooperate separated domains as the application is grow.

2. We must care about which store is injected if multiple stores can be exists. (This is the problem that the single source of truth pattern would like to solve)

3. vue-devtools may not work as expected since it assumes there is only store in each page.

You can use [`namespaced: true` option](https://vuex.vuejs.org/en/modules.html#namespacing) to clearly separate each domain. In addition, we can easily realize which domain the committed mutations belong to on vue-devtools.

yanick:
> It makes sense for testing purpose. However, I think just wrapping by a function is a better solution because we can set various state, actions, getters and mutations for each test case.

That's what I'm doing at the moment, it's a few characters more than the Pr solution, but it works well.

> I think we should use only store even if in such case.

Yeah, I do realize it make the overall architecture simpler if we assume there will ever only be a singleton instance of the store. I however have the feeling that the agglomerate of stores that would be independent will not be as simple as dealing with each individual store. I guess I'll have to have a stab at it and find out. :-)

yyx990803:
I'd just make your modules also export a function, so that you do:

``` js
let new_baz_module = () => ({
    state: { quux: 21 },
    mutations: { 'inner': state => state.quux++ },
});

let new_store = () => new Vuex.Store({
    state: () => { foo: 12 },
    mutations: {
        'bar': state => state.foo++
    },
    modules: { baz: new_baz_module() },
});
```

PR #588, opened 2017-01-17
Update logger.js

marszhou:
add logger a new configurable param 'ignored', let programmer to decide to ignore some log he/she does not want to see.

### DIFF
--- a/src/plugins/logger.js
+++ b/src/plugins/logger.js
@@ -5,6 +5,7 @@ import { deepCopy } from '../util'
 export default function createLogger ({
   collapsed = true,
   transformer = state => state,
+  ignored = type => false,
   mutationTransformer = mut => mut
 } = {}) {
   return store => {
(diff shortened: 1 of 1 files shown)


src/plugins/logger.js
```diff
@@ -5,6 +5,7 @@ import { deepCopy } from '../util'
 export default function createLogger ({
   collapsed = true,
   transformer = state => state,
+  ignored = type => false,
```
ktsn: It would be better to pass mutation object instead of type?

yyx990803: Yeah, I think passing in the entire mutation object leaves room for more flexibility.

ktsn:
@marszhou I'm sorry that we already merged the same filtering feature for logger via #711. We totally forgot this PR.
You can add `filter` option that should have negated return value of your `ignore` option on `createLogger` since Vuex v2.3.1
Thank you for your PR, though.

Issue #598, opened 2017-01-24
Vuex documentation enhancement - return functions in getters


mudrz:
Request for enhancing the documentation

This has been discussed in several other Issues (for example https://github.com/vuejs/vuex/issues/526 and https://github.com/vuejs/vuex/issues/456), and as someone who is learning Vue + Vuex who just went through the same issue, it would be very helpful to have such common cases in the official docs, instead of in the closed issues

Accessing array data is not an isolated use-case, but instead an absolute necessity for larger applications

<details> 
  <summary>Reasoning below (if there is another, better Vuex way, please share it)</summary>
vuex is playing the role from one side as a simple abstraction of the application data

```js
//the state, similar to the DBContext layer in the ORM world
const state = {
    products = [
          { id:1, fullName:'foo', price: 200, color: 'red' },
          { id:2, fullName:'bar', price: 999, color: 'blue' }
    ]
}
//and the abstraction over direct data access, similar to the DBSets in ORMs
const getters = {
    products(state){
      return state.products
    },
    //or simply
   products: state=> 
}
```

But also as a repository pattern substitute, where data changes are done through mutations.
```js
const mutations = {
     //addProduct(state, payload)
     addProduct(state, { fullName, price, color }){
        state.products.push({
            fullName,
            price,
            color,
            id: generateProductId()
        })
     }
}
```
In that sense, I felt that Vuex is in an odd spot (I was comparing Vuex with a standard data repository pattern) where data changes are abstracted within the scope of vuex (mutations), but data getting of array items is supposed to be carried out by a separate application layer, instead of having commonly used getters with arguments:
```js
const dynamicGetters = {
     getProductById: (state, getters, payload) => {
      return getters.products.find(x=> x.id === payload)
    },

     getProduct: (state, getters, { name, color }) => {
      return getters.products.find(x=> x.fullName === name && x.color === color)
    }
}
```
so that queries used from different parts of the application can be cached in a centralized location (the ORM/DB way of thinking, obviously not valid in the Vue + Vuex world) and only specific, non-standard queries are moved to their own layer
</details>

> It is fairly obvious that getters can return functions once you see it, but not so much when you have OOP background

Hopefully this helps others - the upper dynamicGetters (these don't exist in vuex) represented as standard getters that return functions, as suggested by @ktsn :
```js
const getters = {
    //ES6
    getProductById: (state, getters) => (id) =>{
           return getters.products.find(x=> x.id === id)
    },
   //function syntax
[...]

PR #601, opened 2017-01-26
Update getter docs

mudrz:
Add possibility to return functions in getters (https://github.com/vuejs/vuex/issues/598)

### DIFF
--- a/docs/en/getters.md
+++ b/docs/en/getters.md
@@ -61,6 +61,27 @@ computed: {
 }
 ```
 
+You can also return functions in getters:
+```js
+getters: {
+  // ...
+  getTodoById: (state, getters) => (id) => {
+    return getters.todos.find(todo => todo.id === id)
+  }
+  // function syntax
+  getTodoById: function(state, getters){
+    return function(id){
+      return getters.todos.find(todo => todo.id === id)
+    }
+  }
+}
+```
+
+``` js
+store.getters.getTodoById(2) // -> { id: 2, text: '...', done: false }
+```
+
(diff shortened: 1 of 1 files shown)


docs/en/getters.md
```diff
@@ -61,6 +61,27 @@ computed: {
...
 
+You can also return functions in getters:
+```js
+getters: {
+  // ...
+  getTodoById: (state, getters) => (id) => {
+    return getters.todos.find(todo => todo.id === id)
+  }
+  // function syntax
+  getTodoById: function(state, getters){
```
ktsn: Is this necessary? I feel only the arrow function example is enough.

docs/en/getters.md
```diff
@@ -61,6 +61,27 @@ computed: {
 }
 ```
 
+You can also return functions in getters:
```
ktsn: It would be nice if there is a statement that describes this pattern is for passing some arguments to getters. :slightly_smiling_face:

antonstefanov:
Added a bit more details on possible uses to the description

ktsn:
Thanks!

docs/en/getters.md
amirrustam: Honestly, I didn't I know you could do this with getters until I saw this commit. I'm sure many like myself will also benefit from this example.

ktsn: As this example has the same meaning as the above example, I'm still not sure why we need both the arrow function and normal function version in Vuex docs.

PR #629, opened 2017-02-13
Mini fix in rootState example.

agustin-tiendanube:
Maybe I misunderstood the example, but without parenthesis the `%` operator will be executed before the `+`.


ktsn:
@agustin-tiendanube 
Do you mean `(state.count + rootState.count) % 2 === 1`?
`(state.count + rootState.count % 2) === 1` is same as without parenthesis.

agustin-tiendanube:
@ktsn yes that was what i mean!

PR #634, opened 2017-02-14
[docs] add a note that registerModule can receive an array of string

### LINKED ISSUE #633: Documentation on registering modules inside nested modules(namespaced: true)
I was trying to register dynamic module that would be related only to namespaced one and according to [documentation](https://vuex.vuejs.org/en/modules.html) first parameter should be a module name (String). 
But in [sources](https://github.com/vuejs/vuex/blob/dev/src/index.js#L134) I found that it possible to pass Array ['parent', 'myModule'] and 'myModule' will be created inside parent(that was what I needed).
Please update documentation, maybe, it will save someone's time

ktsn:
Fix #633

### DIFF
--- a/docs/en/modules.md
+++ b/docs/en/modules.md
@@ -227,12 +227,18 @@ export function createPlugin (options = {}) {
 You can register a module **after** the store has been created with the `store.registerModule` method:
 
 ``` js
+// register a module `myModule`
 store.registerModule('myModule', {
+  // ...
+})
+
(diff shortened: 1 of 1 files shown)


posva:
I think it'll be better to directly provide an example (in addition to the existing one). Something like:

```js
  // register a nested module nested/module
  store.registerModule(['nested', 'module'], {
    // ...
  })
```

ktsn:
@posva Thanks! I've amended it.

Issue #38, opened 2016-01-13
Am I cheating if I use computed properties with setters?


thelinuxlich:
```
<input v-model="message">

// ...
computed: {
  message: {
    get: function () {
      return store.state.message
    },
    set: function (newValue) {
      store.dispatch("CHANGE_MESSAGE", newValue)
    }
  }
}
//...
```

yyx990803:
Ha, I'd say this is pretty clever. The point is make sure all changes are explicitly recordable, and this does ensure every change to `message` is recorded in a mutation.

blake-newman:
Id say this is a good way to incorporate flux, as long as there is no direct manipulation, so that an action can control calls to multiple mutations. +1

amanpatel:
If this is encouraged, I wonder if it can be made less verbose and perhaps even part of Vuex or Vue natively. I'm thinking of a brand new vuex property declaration:

```js
  vuex: {
    message:  {
      dispatch: "CHANGE_MESSAGE",
      // more configuration?
    }
  }
```

Again, probably a lot of thought needs to go in to this before it is actually incorporated. But I wanted to start the discussion. This also directly addresses a way to to work with v-model on writable computed properties.

This could also be a mixin or plugin opportunity which uses the configuration and creates the computed properties for you. Would love to hear thoughts.

yyx990803:
@amanpatel imo this could be used in certain situations, but I would not put it into core. The point being that this makes it much less explicit about what is going on. It's fine if you build your own abstraction layer on top of Vuex to make this possible (e.g. a mixin), but I think Vuex itself should only provide the most explicit API.

leevigraham:
@thelinuxlich For anyone else interested I had a slightly different requirement where there was lots of child data attributes.

@LinusBorg Helped point me in the right direction over here: https://forum.vuejs.org/topic/1546/vuex-state-mutation-error-on-data-stored-inside-component-data/4

* I'm using a single vue file complied with webpack
* I'm using vuex config over the global store singleton
* Adding getters and setters for each child / form input would have been painful
* The store state is quite deeply nested
* In this case I'm manipulating an object in the store that has children

```
<template>
  <div>
    <h2>Customer Needs Analysis</h2>
    <input 
        :value="customerNeedsAnalysis.vehicleUsage" 
        v-on:input="update('vehicleUsage', $event)"
    >
    <textarea 
        :value="customerNeedsAnalysis.notes" 
        v-on:input="update('notes', $event)"
[...]

PR #645, opened 2017-02-20
Update two-way computed property form example

jblotus:
Added a template example to demonstrate two-way computed bound on the input using v-model.


jblotus:
I was following along with the examples and my original input looked like "<input :value="message" @input="updateMessage">" - this creates an error and I had to look up https://github.com/vuejs/vuex/issues/38 to find the correct usage

Issue #651, opened 2017-02-22
Multiple mutations


Fleeck:
Hi, thanks for your work. Vue & Vuex are really awesome. I was wondering is there any way of commiting multiple mutations at once, passing them as array of objects?

Current code:
```javascript
commit('LOADING', false)
...
(many commits go here)
```

What I was looking for is smth like this:

```javascript
commit([
  {
    type: 'LOADING',
    payload: false
  },
  ...
])
```

Looking forward to your reply about this small issue :)

ktsn:
I like the idea. Also we can compose multiple actions with returned promises.

```js
// current
Promise.all([
  dispatch('FETCH_FOO')
  dispatch('FETCH_BAR')
  dispatch('FETCH_BAZ')
]).then(() => { /* ... */ })

// passing array
dispatch([
  { type: 'FETCH_FOO' },
  { type: 'FETCH_BAR' },
  { type: 'FETCH_BAZ' }
]).then(() => { /* ... */ })
```

Austio:
So I'm not on Vuex core team but am a user of it.

Just from my perspective, this doesn't seem like something that needs to be in vuex b/c it can easily be handled in userland.

BartCorremans:
This seems like it would make more sense for actions (which ktsn is demonstrating) than mutations (which OP is asking for) because actions can return promises. Even then I agree with Austio.

ktsn:
IMHO, it's worth adding if the use case is well common and the users always write such helpers to achieve it. Isn't it common pattern to commit/dispatch batched mutations/actions?

lmiller1990:
Chiming in on this - I have many situations like the OP described, where I do multiple dispatches and wait for them all to resolve before mutating the state. This kind of functionality would be very nice!

posva:
I really think it's not worth adding because in the code you showed:

```js
// current
Promise.all([
  dispatch('FETCH_FOO'),
  dispatch('FETCH_BAR'),
  dispatch('FETCH_BAZ')
]).then(() => { /* ... */ })

// passing array
dispatch([
  { type: 'FETCH_FOO' },
  { type: 'FETCH_BAR' },
  { type: 'FETCH_BAZ' }
]).then(() => { /* ... */ })
```

The first version is explicitly telling you the strategy used for the promise and is pure js so it's knowledge you can reuse elsewhere. The size difference is too small. Also, if I want to add a fetch in the middle of the promise.all, it's much easier with the first solution:

```js
// current
Promise.all([
[...]

PR #666, opened 2017-03-02
Implement #651 multiple mutations

CodinCat:
I implemented the multiple mutations committing feature (#651). So now you can commit multiple mutations with an array.

```js
store.commit([{
  type: 'ADD',
  amount: 1
}, {
  type: 'ADD',
  amount: 2
}])

// equals to
store.commit({
  type: 'ADD',
  amount: 1
})
store.commit({
  type: 'ADD',
  amount: 2
})
```

This will also work: 

```js
store.commit(['INCREMENT', 'INCREMENT', 'INCREMENT'])

// equals to
store.commit('INCREMENT')
store.commit('INCREMENT')
store.commit('INCREMENT')
```

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -293,6 +293,16 @@ function makeLocalContext (store, namespace, path) {
     },
 
     commit: noNamespace ? store.commit : (_type, _payload, _options) => {
+      if (Array.isArray(_type)) {
+        _type.forEach(mutation => {
+          if (typeof mutation === 'string') {
+            local.commit(mutation, null, _payload)
+          } else {
+            local.commit(mutation, _payload)
+          }
+        })
+        return
+      }
       const args = unifyObjectStyle(_type, _payload, _options)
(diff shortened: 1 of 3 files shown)


ktsn:
In addition, I think `commit` should receive options object as 2nd argument if the 1st argument is an array. So that we can commit mutations in global namespace.

```js
actions: {
  foo ({ commit }) {
    commit([
      { type: 'TEST', payload: 1 },
      { type: 'TEST', payload: 2 }
    ], { root: true })
  }
}
```

CodinCat:
Ok updated the code. Not sure it looks a little bit dirty or just fine.

thanks

ktsn:
According to original thread reactions and since this can be easily achievable without modifying Vuex, I guess this should not be included into Vuex. I'm sorry to make it back and forth but thank you for your PR anyways!

PR #698, opened 2017-03-19
[improvement] Add assertion to verify using of new

termosa:
I see people are asking about this problem so often, that I feel it worth to put this assertion into the code to save some time for those developers.


termosa:
@ktsn sure, it's [here](https://github.com/vuejs/vuex/pull/698/commits/7d2c3e30531bace1cdbc92062d3703cf98aa5716)

PR #705, opened 2017-03-23
Add assertion of module assets

### LINKED ISSUE #285: Bug in vuex & vuex-router-sync: rawGetter is not a function
When click the route-link to another component, this error will show. I found it happen at vuex 2.0.0-rc.5 dist/index.js line 458 "store._wrappedGetters[**getterKey**]". When the getterKey is "default", this is an Object, not a Function. I don't know it is my problem or the bug.

ktsn:
Fix #285 

Sometimes the users can accidentally pass wrong values to getters/actions/mutations options. It causes an unexpected and unpredictable error like reported on #285.
This patch add assertions for them and let the users know which value is problematic in early timing.


yyx990803:
Hmm, maybe the assertions could be dev-only like in vue core / vue-router?

ktsn:
@yyx990803 Updated the build script based on vue-router's one and added env check for all assertions/warnings.

yyx990803:
👍 nice job

PR #707, opened 2017-03-24
Fix minor code error in store option example

hmans:
Just a small thing I stumbled over.


BartCorremans:
This is not an error, it's valid ES6 syntax: http://es6-features.org/#PropertyShorthand

ktsn:
Thanks for the PR.
As @BartCorremans is described, this is valid ES6 syntax.

hmans:
Duh. That's what I get for spending too much time in ES5 land. Thanks!

PR #711, opened 2017-03-25
Add filtering of the mutations logging

### LINKED ISSUE #710: Add filter to the logger
In my application, I need to display the current time, so I added a field that I update each second with a `new Date()`. The trouble is that it floods the console with updates notifications. I would like to be able to log all the mutations except this one.

```
const logger = createLogger({
    filter(type, state_before, state_after) {
        return type !== "update_time"
    },
})
```
I personally only care about `type`, but if `state_before` and `state_after` are not too complicated to compute it may be useful.

Yamakaky:
Fixes #710

### DIFF
--- a/docs/en/plugins.md
+++ b/docs/en/plugins.md
@@ -102,7 +102,7 @@ The `createLogger` function takes a few options:
 ``` js
 const logger = createLogger({
   collapsed: false, // auto-expand logged mutations
-  filter(mutation, stateBefore, stateAfter) {
+  filter (mutation, stateBefore, stateAfter) {
     // returns true if a mutation should be logged
     // `mutation` is a { type, payload }
     return mutation.type !== "aBlacklistedMutation"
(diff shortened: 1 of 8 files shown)


docs/en/plugins.md
```diff
@@ -102,6 +102,11 @@ The `createLogger` function takes a few options:
 ``` js
 const logger = createLogger({
   collapsed: false, // auto-expand logged mutations
+  filter(mutation, stateBefore, stateAfter) {
```
ktsn: Please add a space before the open parenthesis 🙂

erfe1: Fr

Yamakaky: ?

Yamakaky:
Anything missing before merge?

ktsn:
Waiting for the review by @yyx990803 :slightly_smiling_face:

Yamakaky:
OK

Yamakaky:
Do you have the time to do the review @yyx990803 ?

Yamakaky:
Cool! When do you think it will be released?

Yamakaky:
Any release date idea?

Issue #715, opened 2017-03-27
Observable for dispatch()?


aicest:
### What problem does this feature solve?
Allow return of Observable by `dispatch()`

### What does the proposed API look like?
```
Vue.use(Vuex, Observable)

const observable = store.dispatch('observable-action') // Observable<any> => Observable<any>
const promise = store.dispatch('promise-action') // Promise<any> => Promise<any>
const promise = store.dispatch('normal-action') // any => Promise<any>

const observable = store.dispatch('multi-observable-action') // Observable<any>[] => Observable<any[]>
const observable = store.dispatch('multi-observable-or-promise-action') // (Observable<any>|Promise<any>)[] => Observable<any>[] => Observable<any[]>
const promise = store.dispatch('multi-promise-action') // Promise<any>[] => Promise<any[]>
const promise = store.dispatch('multi-normal-action') // any[] => Promise<any>[] => Promise<any[]>
```

aicest:
@ktsn PR #716

ktsn:
Thank you for your contribution, however IMO this should be implemented as a plugin.

If we support various data type of async control flow, the code base would be bloated and it would be less maintainable since we always need to consider about all async formats in the future.
In addition AFAIK, the most of Observable implementations have `fromPromise` helper to generate an Observable instance from a Promise object. So you already create it easily without modifying Vuex code.

LinusBorg:
Agree with ktsn, but I like the idea. Should definitely be a plugin

ktsn:
Actually, it would be so easy to create such plugin in user land - just wrap `dispatch` and convert returned `Promise` to `Observable`. So this would not need to be in Vuex core. Thanks for your suggestion, though!

```js
const observablePlugin = (Observable) => (store) => {
  const dispatch = store.dispatch
  store.dispatch = function observableDispatch(...args) {
    return Observable.fromPromise(dispatch(...args))
  }
}

const store = Vuex.Store({
  // ...
  plugins: [
    observablePlugin(Observable)
  ]
})
```

PR #716, opened 2017-03-27
#715: Allow return of Observable by dispatch()

aicest:


ktsn:
See https://github.com/vuejs/vuex/issues/715#issuecomment-329525479

PR #723, opened 2017-03-31
Pass original mutation to subscribers on object-style commit

Szellem:
After object-style commit subscribers got back the original mutation
wrapped into a new mutation as payload. This was counter-intuitive
and prevented the reuse of mutation objects. For example after the call
commit({ type: 'TEST', amount: 2 }) one would expect to receive the
same object in the subscriber but actually
{ type: 'TEST', payload: { type: 'TEST', amount: 2 } })
was received with the original implementation.

### DIFF
--- a/dist/vuex.js
+++ b/dist/vuex.js
@@ -311,8 +311,8 @@ Store.prototype.commit = function commit (_type, _payload, _options) {
     var type = ref.type;
     var payload = ref.payload;
     var options = ref.options;
+    var mutation = ref.mutation;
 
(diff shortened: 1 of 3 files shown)


ktsn:
Sounds reasonable. However, this seems breaking change...
@yyx990803 What do you think?

yyx990803:
Thanks for the PR. This is actually intended because this ensures plugins (library, static code) always receive consistent data format regardless of what the commit style the user decides to use (userland, dynamic code).

While I think it's nice in some way, I don't think it's worth making a breaking change for it.

PR #733, opened 2017-04-07
Improve module getters definition

Healkiss:
It missing rootGetters for the getters definition in API.


roblav96:
@Healkiss Yes please merge =]

PR #743, opened 2017-04-18
describe options argument of dispatch/commit in api docs

### LINKED ISSUE #742: API documentation doesn't mention options argument for commit
API misses that commit() call can have third options parameter, that can contain very important option 'root: true', making it hard to find how to commit mutations to the root store.

For the API documentation page:
commit(type: string, payload?: any, options?: object) | commit(mutation: Object)

And some note should be added to 'mutations' section of the documentation.

ktsn:
Fix #742


ktsn:
@posva Thanks!

PR #761, opened 2017-04-29
fix .toggle-all checkbox was hidden

huguangju:
see tastejs/todomvc-app-css#8.

`todomvc-app-css` upgraded version to `2.1.0`, `"todomvc-app-css": "^2.0.6"` in vuex's package.json will download latest version, `.toggle-all` class has a big change, causing it to be hidden in todomvc example.

### DIFF
--- a/examples/todomvc/components/App.vue
+++ b/examples/todomvc/components/App.vue
@@ -13,10 +13,11 @@
     </header>
     <!-- main section -->
     <section class="main" v-show="todos.length">
-      <input class="toggle-all"
+      <input class="toggle-all" id="toggle-all"
         type="checkbox"
         :checked="allChecked"
         @change="toggleAll({ done: !allChecked })">
+      <label for="toggle-all"></label>
       <ul class="todo-list">
         <todo v-for="(todo, index) in filteredTodos" :key="index" :todo="todo"></todo>
       </ul>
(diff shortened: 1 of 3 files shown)


examples/todomvc/components/App.vue
```diff
@@ -13,10 +13,11 @@
     </header>
     <!-- main section -->
     <section class="main" v-show="todos.length">
-      <input class="toggle-all"
+      <input class="toggle-all" id="toggle-all"
```
posva: where does the id comes from? It's not on the todomvc repo

huguangju: Yes, it's not. The checkbox arrow(`.toggle-all`) need a label tag next to it, and can't be wrapped with label, because it was written like this in todomvc-app-css:
```css
.toggle-all + label:before {
    content: '❯';
    font-size: 22px;
    color: #e6e6e6;
    padding: 10px 27px 10px 27px;
}
```
Then if i want to click the arrow and trigger the event, need to bind the `label` tag to input with the `for` attribute, so i set id on input. Is there any other way better than this?

posva: I see, yeah the `+` doesn't leave us other option here (at least none that I'm aware of)

ktsn:
Thanks!

PR #763, opened 2017-05-01
Add logo to docs

marshallswain:
This project deserves its own logo.  This adds a 2kb svg file to each language in the vuex@2.x docs.

Preview:
screen shot 2017-05-01 at 7 45 14 am


posva:
Good job, although, I think it'd be better if the logo reminds of the cycle as the redux logo does

marshallswain:
@posva  Thanks.  I thought about that.  I'll see if I can work something in.

phanan:
Not a designer, but I think the width-height proportion of the logo is a bit… strange. A more square-ish logo would fit better for other purposes e.g. favicons and all.

marshallswain:
Here's a new logo revision.  I tried a bunch of combinations of orbits. Two orbits stays the cleanest without ruining the `X`symbol.  (Adding more starts to make it feel like a hazmat symbol due to the sharp edges inside the circular areas ;)  The rings also help to square it up a bit, though it was only a few pixels off square, before.  For smaller sizes, like a favicon, I would probably stick to one, thicker orbit.  I also brought the central diamond shape to the front of the orbit rings to emphasize the embedded Vue logo and the X, underneath.

screen shot 2017-05-02 at 12 40 14 am

posva:
What does it look like if you remove the bottom part and center the orbits
again?

marshallswain:
No X, Centered Orbits
no-x-centered-atoms

No X, Centered Orbits, Darker Orbit
no-x-centered-atoms copy


Lighter X, Centered Orbits
centered-atoms-lighter-x copy


Centered Orbits:
centered-atoms-lighter-x

anteriovieira:
**No X, Centered Orbits, Darker Orbit** 👍

marshallswain:
Heh.  The X isn't really needed, anymore.  It's in the orbits.  This is why I'm not a designer.  I'm a hobbyist.  😉

yyx990803:
Hey @marshallswain - thanks a lot for the proposal! I actually thought of designing logos for sub-projects too, but ended up not really doing it because vuex and vue-router are both tightly coupled to Vue and I'd rather just use a single logo to represent the entire Vue stack :) Plus it's also hard to get it right (annnnd we'd need one for vue-router too if we have one for vuex?) But still, really appreciate your effort!

Issue #467, opened 2016-11-19
Enable Plugins in Modules


ghost:
Currently you cannot define a plugin when using a module store. I would like to be able to define plugins as part of a module that is being registered. I have built a `user` store that accepts the id, email and key for the user. I would like to define a plugin that saves the email to build remember me functionality.

```
store.registerModule('user', {
  state: {
    id: null,
    email: window.localStorage.getItem('email') || null,
    key: null
  },

  mutations: {
    register: (state, payload) => state.email = payload.email
  },

  plugins: [
    (store) => {
      store.subscribe((mutation, state) => {
        if (mutation.type === 'register') {
          window.localStorage.setItem('email', state.email)
        }
    }
  ]
})
```

ktsn:
Sounds good to me 🙂 
One thing we may need to consider is how to unregister in-module plugins. As modules can be unregistered dynamically, in-module plugins can be also unregistered, I think.

yanick:
For what it's worth, I have a first stab at that at https://github.com/yanick/vuex/tree/plugins-in-modules

Since a plugin can subscribe/watch to many things, I assume that plugins that want to be able to unregister as their module will be have to return an anonymous function that will call the unregistering function of all those subscribes/watches.

Of course, if the plugin only does a single subscribe, like
```
plugins: {
   foo: store => store.subscribe( ... )
}
```

it'll all magically work (as the return value of the subscribe is passed directly as the return value of the plugin.

snaptopixel:
I'm not sure "plugins" are appropriate for this type of behavior (I actually found this issue trying to implement a similar strategy as @JBodkin).

It may be better to have lifecycle hooks instead:

```js
const myModule = {
  mutations: …
  actions: …
  registered(store) {
    getSomethingFromSomewhere().then(thing => store.commit('GOT_THING', thing));
    this.unsubscribe = store.subscribe(…);
  }
  unregistered(store) {
    this.unsubscribe();
  }
```

snaptopixel:
FWIW, in my project I just did this in the store:
```js
plugins: [
  store => store.dispatch('STORE_INIT')
],
```
Then any modules can have a `STORE_INIT` action that can do whatever setup needed. In my case it was calling `firebase.auth().onAuthStateChanged` to check for existing auth.

yanick:
> Then any modules can have a STORE_INIT action that can do whatever setup needed.

True. Of course, the actions don't pass the store itself as the context, but that can be worked around:

```
plugins: [
[...]

PR #773, opened 2017-05-10
Enable plugins in modules

crossjs:
fix: #467.

Plugins in modules now have a localized subscribe.
Mutations in children will fire parents' subscriptions too.

### DIFF
--- a/dist/vuex.js
+++ b/dist/vuex.js
@@ -1,5 +1,5 @@
 /**
- * vuex v2.3.0
+ * vuex v2.3.1
  * (c) 2017 Evan You
  * @license MIT
  */
@@ -139,6 +139,9 @@ Module.prototype.update = function update (rawModule) {
   if (rawModule.getters) {
     this._rawModule.getters = rawModule.getters;
   }
+  if (rawModule.plugins) {
+    this._rawModule.plugins = rawModule.plugins;
+  }
 };
 
 Module.prototype.forEachChild = function forEachChild (fn) {
@@ -160,6 +163,12 @@ Module.prototype.forEachAction = function forEachAction (fn) {
 Module.prototype.forEachMutation = function forEachMutation (fn) {
   if (this._rawModule.mutations) {
     forEachValue(this._rawModule.mutations, fn);
+  }
+};
+
+Module.prototype.forEachPlugin = function forEachPlugin (fn) {
+  if (this._rawModule.plugins) {
+    forEachValue(this._rawModule.plugins, fn);
   }
 };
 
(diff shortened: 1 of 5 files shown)


ktsn:
Thank you for your contribution! Before diving into the code details, I would like to discuss about the API.

At first, do we really need plugins in modules? As far as I can see, people would really need `registered` lifecycle to execute initial actions/subscribe. It would be better to implement minimal feature that satisfy the demands. I think we need to simulate the root store by local context as possible if we really support plugins in modules because any 3rd party plugins can be used for it. Then it would be harder to maintain Vuex.

About subscribe method in local context, should we filter the execution of callback by namespace? As the whole state is shaped as tree structure, parent module may mutate child state and it wouldn't work well in that case.

Also, it would be great if we can pass a module's local state to subscribed callbacks. 🙂

ktsn:
What do you think, @yyx990803 ?

crossjs:
For example, I have two modules which have no namespaces declared:

module A:
```js
export default {
  namespaced: true, state, getters, actions, mutations, plugins
}
```

and module B:
```js
export default {
  namespaced: true, state, getters, actions, mutations, plugins
}
```

put them to one store:
```js
import moduleA from 'moduleA'
import moduleB from 'moduleB'

new Vuex.Store({
  state,
  // Namespaces could be customized, even if a random string.
  // And in the case, we also inject namespace to modules' state/getters/actions/mutations.
  // So, in moduleA and moduleB, we can use them like there is NO namespace.
  modules: {
    nsX: moduleA,
    nsY: moduleB
  }
})
```

ktsn:
I see. But why does this have to be a plugin? I'm still not sure about that.

Sydtrack:
I have one specific use case where module plugins would be an excellent feature: it is a mobile app that must work offline and sync back to server when a connection is available. Users must be able to select any of the companies assets and register some issues on them. The problem is that I have about 30k assets and they must be available offline. 

When users are online, I download the complete assets data to a local cache. I use a modified vuex-persist plugin for storing this state on disk. It is modified because Vuex-persist works by dumping the complete state after each commit. Since his module data will never change outside a sync action, there should be no reason to dump it every time. 

I changed vuex-persist so I could pass it the module namespace, allowing me to save different files (or storage keys) for each module. This way I can include multiple copies of the plugin, each for a different module I wish to persist. The namespace also allows me to ignore mutations that aren't present on the referenced module, reducing drastically what needs to be dumped. 

Module plugins would make all this trouble a lot easier.

Well, in a quick TL;DR: +1

kiaking:
Closing due to inactivity. The PR is too old to be merged I think 😓

PR #776, opened 2017-05-12
Add a try-catch to the testAction helper method example

Mark-de-Haan:
When an error occurs in an assertion, you receive the message that your promise hook was not closed within 2 seconds.
By catching the error, we can call done(error) (this works for standard vue-cli template)
1. This makes sure the test is properly closed
2. The assertion error is logged to the console

### DIFF
--- a/docs/en/testing.md
+++ b/docs/en/testing.md
@@ -92,10 +92,16 @@ const testAction = (action, payload, state, expectedMutations, done) => {
   // mock commit
   const commit = (type, payload) => {
     const mutation = expectedMutations[count]
-    expect(mutation.type).to.equal(type)
-    if (payload) {
-      expect(mutation.payload).to.deep.equal(payload)
+    
+    try {
+      expect(mutation.type).to.equal(type)
+      if (payload) {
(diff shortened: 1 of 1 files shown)


posva:
what test framework are you using? A non-caught exception in the synchronous code should make the test fail

Mark-de-Haan:
Im using Karma + Mocha (from the standard vue init command)
Im testing my actions, and mocked the response of any API calls using testdouble.

I am using the `testAction` helper function found in the Vuex docs.
In case of expect(1).to.equal(2) inside the the testAction function, you will only get the message that the Promise hook has not been closed. 

You will **not** get the notification that your expected value 1 does not equal 2

I could try-catch the actual method call itself, but seeing as I call the testAction function very often, Id rather have it at the spot where my actual expect assertions are being done

posva:
Aaah, I see. I didn't see there was a helper there calling the mutation in a callback. I prefer not mocking everything when testing actions and dispatch them, this way I get a Promise and I can append `.catch(done)`.

I think that in this case the try/catch is ok 🙂

PR #797, opened 2017-05-23
Update intro.md

gavenheim:
Adds space between `width: 100%` and `max-width: 450px` so that the one way data flow image doesn't get cut off on mobile devices.


ktsn:
@gavenheim 
Is it still cut off on mobile? I guess it was not deployed yet on that time, it's now updated.
https://vuex.vuejs.org/en/intro.html

The space in style attribute seem not to have an effect since the properties are separated by semi-colons.

gavenheim:
@ktsn 
Nope, looks like the docs are no up to date now. I thought something weird was going on. Anyway, closing this PR.

PR #820, opened 2017-06-05
Doc EN: Webpack to webpack

Haeresis:
- No usage of lower capital to « webpack » word fix (even as starting word
   - — see https://webpack.js.org/configuration/ for phrase majuscule example 
   - — see https://webpack.js.org/development/release-process/ for link/title example)


ChalkPE:
@kazupon #835 :tada:

PR #834, opened 2017-06-15
add Store#hasModule(path) API

FranckFreiburger:


src/store.js
```diff
@@ -179,6 +179,16 @@ export class Store {
...
   }
 
+  hasModule (path) {
+    if (typeof path === 'string') path = [path]
+
+    if (process.env.NODE_ENV !== 'production') {
+      assert(Array.isArray(path), `module path must be a string or an Array.`)
+    }
+
+    this._modules.isRegistered(path)
```
ktsn: We must return the result of this expression.

src/module/module-collection.js
```diff
@@ -53,6 +53,14 @@ export default class ModuleCollection {
 
     parent.removeChild(key)
   }
+
+  isRegistered (path) {
+    const parent = this.get(path.slice(0, -1))
```
ktsn: I realized this line can throw an error if the parent module does not exist.
To handle this, we also need to add existance check in `get` method.

FranckFreiburger: yes,
```
store.registerModule('hi' ...
expect(store.hasModule('hi')).toBe(true)
expect(store.hasModule(['hi','big'])).toBe(false)
expect(store.hasModule(['hi','big', 'world'])).toBe(false) // THROW
expect(store.hasModule(['bonjour'])).toBe(false)

```

FranckFreiburger: `store.registerModule(['aze', 'rty'], ... // THROW`

FranckFreiburger: maybe this behavior is expected, maybe it's a documentation issue ?

ktsn: But the purpose of this feature is to check existance of module, isn't it? I think the `hasModule` should return `false` in that case.

FranckFreiburger: yes, this feature is to check existence of module, but maybe not the existence of sub-module.  
It's quite like `mkdir` and `mkdirp`

FranckFreiburger: I propose to add `ModuleCollection#has(path)`

ktsn: I'm not sure why this method is semantically related with `mkdir`. They are used on different platform (Web/CLI) and also are different kind of task (`hasModule` *checks* a module while `mkdir` *creates* a directory).

If the parent module does not exist, it's clear that its child module also does not exist. So why `hasModule` need to throw in that case? 🤔

ktsn:
* Can we add **a new test case** for `hasModule`? Adding a test code to non-relevant test case would problematic since it would make harder to detect a real problem when the test cases fail.
* Looks like CI is failed due to linting issue. Please check/fix it by running `npm test` on your terminal.

src/module/module-collection.js
FranckFreiburger: Trust me, I am not trying to create modules with mkdir :smiley:
I just wanted to say that the responsibility lies with the caller to ensure the correctness of the arguments.

FranckFreiburger:
At this point, `hasModule()` behaves like `register()` and `unregister()` in term of error management, it throws if parent module does not exist.
I will propose a new PR for existence check in get method and update these 3 methods.

anteriovieira:
What prevents it from being merged?

trainiac:
@ktsn Can this be merged? Or does this PR need more work?

mlanin:
@ktsn a year has passed. Can we merge this already? Please :)

wa3l:
@yyx990803 any timeline on merging this?

mlavina:
@ktsn @yyx990803 This would also be really helpful to my team can we get this merged in and released?

JasonLandbridge:
2020 now, this is such a small pull request. Just merge it already!

bhaskar-nair2:
Waiting for this to be available in master...

Issue #789, opened 2017-05-19
`registerModule` resets the sub-state of the module being registered


sirlancelot:
### Version
2.3.0

### Reproduction link
[https://codepen.io/sirlancelot/pen/EmOxWZ](https://codepen.io/sirlancelot/pen/EmOxWZ)

### Steps to reproduce
Suppose you have a SPA-style app with Vue, Vuex, and Vue-Router, with a Vuex module for each page. The Vuex module gets registered during `router.beforeResolve` on both server & client.

Both the server & client will render the page properly. However, when loading the page fresh in the browser, it seems that `registerModule` is resetting the module's sub-state upon registration.

### What is expected?
When Vuex is registering modules, it should see if that sub-state already exists in its store and use that instead of starting fresh.

### What is actually happening?
It seems that the module's sub-state is getting cleared during `registerModule` phase in the browser.

---
I hacked together the repro steps without Vue-Router because it's not strictly part of the problem. I tried to explain at what point portions of code run. Let me know if you need more explanation.

ktsn:
Maybe we can add a third argument of flag to prevent override the existing state for hydration?

```
registerModule(path: string | Array<string>, module: Module, hydration?: boolean)
```

sirlancelot:
^ Sounds good to me!

I currently do something like this as a workaround:

```javascript
function beforeRouteEnter(to, from, next) {
	patchDefinitionWithCurrentState(definition, store.state[namespace])
	store.registerModule(namespace, definition)
}

function patchDefinitionWithCurrentState(definition, currentState) {
	if (!currentState) return

	const originalState = definition.state

	// Vuex will call this function which will return the definition back to its
	// original state but will provide the currentState back to Vuex.
	definition.state = () => {
		definition.state = originalState
		return currentState
	}
}
```

mAAdhaTTah:
Ran into this issue myself. Would be really cool if the module's `state` (could) provide defaults for any undefined values.

sirlancelot:
I revised my function to provide defaults like so:

```javascript
function patchDefinitionWithCurrentState(definition, currentState) {
	if (!currentState) return

	const state = definition.state

	// Vuex will call this function which will return the definition back to its
	// original state but will provide the currentState back to Vuex.
	definition.state = () => {
		const defaultState = typeof state === "function" ? state() : state
		definition.state = state
		return { ...defaultState, ...currentState }
	}
}
```

mAAdhaTTah:
[...]

PR #837, opened 2017-06-22
feat: preserve state with registerModule

blake-newman:
As speaking with @TheLarkInn we where saying it would be awesome to be able to register a module dynamically to be able to code split the store in a large application. 

It is totally possible with `store.registerModule('a', module)`. Here is an example, if using hooks to ensure prevent the route loading. This can be achieved with `asyncData` with client mixins and global hooks. https://ssr.vuejs.org/en/data.html

```js
export default {
  async asyncData ({ registerModule, dispatch }) {
    registerModule('a', await import('@/store/modules/a'))
    dispatch('a/action')
  }
}
```

The issue with this with SSR this would replace the pre-populated data. This pull request aims to fix this scenario.

### DIFF
--- a/docs/en/api.md
+++ b/docs/en/api.md
@@ -156,9 +156,11 @@ const store = new Vuex.Store({ ...options })
 
   Most commonly used in plugins. [Details](plugins.md)
 
-- **`registerModule(path: string | Array<string>, module: Module)`**
+- **`registerModule(path: string | Array<string>, module: Module, options?: Object)`**
 
   Register a dynamic module. [Details](modules.md#dynamic-module-registration)
+
+  `options` can have `preserveState: true` that allows to preserve the previous state. Useful with Server Side Rendering.
 
 - **`unregisterModule(path: string | Array<string>)`**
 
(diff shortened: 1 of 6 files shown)


ktsn:
This would fix #789

blake-newman:
Awesome, had this in my mind for months just forgot about it.

TheLarkInn:
does this mean you could pass the import directly also?

`this.$store.replaceModule('a', import('@/store/modules/a'))`

blake-newman:
Not yet! This is my next thing to do. Small things at a time haha!

TheLarkInn:
😎 np cheers on this addition for SSR

andreiglingeanu:
🤘

blake-newman:
@TheLarkInn I'm going to go with the latter, as it will require less code in Vuex and a simple integration. Also doesn't require the need to do custom waiting mechanisms for promise to resolve.

TheLarkInn:
So like this?

`this.$store.replaceModule('a', () => import('@/store/modules/a'))`

blake-newman:
yes but it is `registerModule`

PR #839, opened 2017-06-23
Feature: allow same paradigm for lazy loading modules

blake-newman:
As described in https://github.com/vuejs/vuex/pull/837

This will keep the same paradigm in Vuex as rest of ecosystem for dynamically loading store modules:

`store.registerModule('a', () => import('@/store/modules/a'))`

### DIFF
--- a/docs/en/modules.md
+++ b/docs/en/modules.md
@@ -244,6 +244,58 @@ Dynamic module registration makes it possible for other Vue plugins to also leve
 
 You can also remove a dynamically registered module with `store.unregisterModule(moduleName)`. Note you cannot remove static modules (declared at store creation) with this method.
 
+### Lazy Load Dynamic Modules
+
+You can code split you Vuex store and dynamically register store modules. 
+
+``` js
+const myModule = () => import('path/to/module')
+
+store.registerModule('myModule', myModule)
+```
+
+This is simple enough, however you will need to ensure that modules are registered before you can use them. Attempting to use a module before it is registered will not work and will cause errors. 
+
+For example, with vue-router, you can ensure that routing does not resolve until the required modules have been loaded. You can find further [implementation details here](https://ssr.vuejs.org/en/data.html) under the **Client Data Fetching** section. Using this pattern will ensure that we can register modules before use within components.
+
+Overall this pattern will require more awareness and management of the store module registration, however if well managed it can be a very poweful feature and significantly reduce the initial bundle size of your application.
+
+``` js
+import { mapGetters } from 'vuex'
+
+const myModule = () => import('path/to/module')
+
+export default {
+  asyncData(store) {
+    return store.registerModule('myModule', myModule)
+      .then(() => store.dispatch('myModule/getData'))
+  },
+
+  computed: {
+    ...mapGetters('myModule', ['a'])
+  }
+}
+```
+
+Using async and await this becomes less verbose for the registering of a module:
+
+``` js
+import { mapGetters } from 'vuex'
+
+const myModule = () => import('path/to/module')
+
+export default {
+  async asyncData(store) {
+    await store.registerModule('myModule', myModule)
+    await store.dispatch('myModule/getData')
+  },
+
+  computed: {
+    ...mapGetters('myModule', ['a'])
+  }
+}
+```
+
 ### Module Reuse
 
 Sometimes we may need to create multiple instances of a module, for example:
--- a/src/store.js
+++ b/src/store.js
@@ -158,10 +158,15 @@ export class Store {
       assert(path.length > 0, 'cannot register the root module by using registerModule.')
     }
 
-    this._modules.register(path, rawModule)
-    installModule(this, this.state, path, this._modules.get(path))
-    // reset store to update getters...
-    resetStoreVM(this, this.state)
+    const moduleResolved = module => {
+      this._modules.register(path, module.default || module)
+      installModule(this, this.state, path, this._modules.get(path))
+      // reset store to update getters...
+      resetStoreVM(this, this.state)
+    }
+
+    if (typeof rawModule === 'function') return rawModule().then(moduleResolved)
+    moduleResolved(rawModule)
   }
 
   unregisterModule (path) {
--- a/test/unit/modules.spec.js
+++ b/test/unit/modules.spec.js
@@ -82,6 +82,36 @@ describe('Modules', () => {
     })
   })
 
+  it('dynamic module registration with promise (dynamic import)', () => {
+    const store = new Vuex.Store({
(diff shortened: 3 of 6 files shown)


src/store.js
```diff
@@ -158,10 +158,16 @@ export class Store {
...
-    this._modules.register(path, rawModule)
-    installModule(this, this.state, path, this._modules.get(path))
-    // reset store to update getters...
-    resetStoreVM(this, this.state)
+    const moduleResolved = module => {
+      this._modules.register(path, module.default || module)
+      installModule(this, this.state, path, this._modules.get(path))
+        // reset store to update getters...
+      resetStoreVM(this, this.state)
+      return module
```
ktsn: Why do we need to return the module here? I have no idea about use cases 🤔

src/store.js
```diff
@@ -158,10 +158,16 @@ export class Store {
...
+    }
+
+    if (isPromise(rawModule)) return rawModule.then(moduleResolved)
+    return moduleResolved(rawModule)
```
ktsn: It would be better to always return Promise for consistency?

ktsn:
I have a small concern of this API. When some getters/actions/mutations are used before loading the corresponding module, it can be failed. 
I guess the returning Promise of `registerModule` is there for handling such situation. But I wonder if this syntax is needed for Vuex since we need to handle the loading completions of async modules in the user land. Or are you considering like lazy dispatch/commit to handle it in Vuex?

test/unit/modules.spec.js
```diff
@@ -82,6 +82,35 @@ describe('Modules', () => {
...
+    const actionSpy = jasmine.createSpy()
+    const mutationSpy = jasmine.createSpy()
+    const module = () => Promise.resolve({
+      state: { value: 1 },
+      getters: { foo: state => state.value },
+      actions: { foo: actionSpy },
+      mutations: { foo: mutationSpy }
+    })
+
+    store.registerModule(['a', 'b'], module()).then(() => {
```
blake-newman: Yes good spot, either way it will work

src/store.js
blake-newman: If you return a Promise, it will mean you will need to ensure that Vuex is used with Promises which may require pollyfils. This will ensure users don't need it if they are following standard flow

blake-newman:
@ktsn I will run through the code and check some of points, some points may be right to change need to explore. 

The main purpose of the API is to keep code splitting of modules simple to use and consistent with other parts of the Vue ecosystem.

Perhaps extra documentation for lazy loading is needed to make this feature more informative. I think having a section on Lazy loading dynamic modules would be great. 

There is some gotchas users should be aware of, using the client `asyncData` mixins from SSR docs makes this easier. I was thinking that the client mixins are actually more fitting to be in main Vue docs as it can be used without SSR. 

None the less we should highlight prerequisites and patterns required for this use, in the right environment this feature will be extremely powerful for applications that are large.

src/store.js
ktsn: [Vuex requires Promise to work](https://github.com/vuejs/vuex/blob/dev/src/store.js#L12). So we can consider that Promise is always there 🙂

blake-newman:
Furthermore doing the latter of your comment would invoke technical complexity, as this mains if you use an action to get changes is will be deferred. 

The current API just makes it easier to use code splitting but will require user to correctly manage registering modules.

ktsn:
@blake-newman 
I see the point. I agree that the extra documentation for lazy loading would be informative 🙂

nickmessing:
@blake-newman, thank you for the effort, was thinking about this feature for some time, would help me a lot.

types/index.d.ts
```diff
@@ -22,6 +22,7 @@ export declare class Store<S> {
 
   registerModule<T>(path: string, module: Module<T, S>): void;
   registerModule<T>(path: string[], module: Module<T, S>): void;
+  registerModule<T>(path: string[], module: () => Promise<Module<T, S>>): Promise<Module<T, S>>;
```
ktsn: Looks like it returns `Promise<void>`

docs/en/modules.md
```diff
@@ -242,7 +242,57 @@ The module's state will be exposed as `store.state.myModule` and `store.state.ne
 
 Dynamic module registration makes it possible for other Vue plugins to also leverage Vuex for state management by attaching a module to the application's store. For example, the [`vuex-router-sync`](https://github.com/vuejs/vuex-router-sync) library integrates vue-router with vuex by managing the application's route state in a dynamically attached module.
 
-You can also remove a dynamically registered module with `store.unregisterModule(moduleName)`. Note you cannot remove static modules (declared at store creation) with this method.
```
ktsn: Why did you remove this? 🤔

types/index.d.ts
blake-newman: If you don't return it will return the previous resolve value

blake-newman: Which will be the module

docs/en/modules.md
blake-newman: Looks to be an accident... Will ammend

yyx990803:
I like the idea but I'm not sure if this is entirely necessary, because `reigsterModule` is always used imperatively, you can already do:

``` js
asyncData (store) {
  return import('./foo.module.js')
    .then(m => store.registerModule('foo', m))
    .then(() => store.dispatch('foo/bar'))
}
```

Or

``` js
async asyncData (store) {
  store.registerModule('foo', await import('./foo.module.js'))
  await store.dispatch('foo/bar')
}
```

The primary reason for the async component syntax is because components need to be used in declarative scenarios where `await` or Promise resolving are not available. If we were to support async vuex modules with the same syntax, it makes more sense to support it when declaring stores:

``` js
  modules: {
    foo: import('./foo.module.js')
  }
})
```

But this is extremely tricky to get right, since the state access is synchronous.

blake-newman:
@yyx990803 agree. Probably just needs documentation rather than internal changes. IMO in hindsight this doesn't add much value other than syntax sugar that only abstracts implementation detail that is truly not that much more difficult.

alexjoverm:
I agree on this is not necessary, but I consider that if it doesn't cause any trouble, it's a nice syntax sugar added.

When writing [Lazy load in Vue using Webpack's code splitting](https://alexjoverm.github.io/2017/07/16/Lazy-load-in-Vue-using-Webpack-s-code-splitting/), I could easily realize it's easy by doing it in the way Evan mentioned. But some people were expecting it to work in the same way than components or router (even some though `registerModule` understand import function already).

Aside: does it make sense to lazy load modules on declaration? I think the use case is to load them under a determined condition

PR #843, opened 2017-06-28
add doc complement about on the fly module creation

simdax:
Hello, 

I was fighting to create on the fly module components lists, but I believe I succeded.
I wrote a little addendum for the documentation.

Maybe it could be usefull ?


blake-newman:
`state` can  be a function, which can return a new object. As described here https://vuex.vuejs.org/en/modules.html under the `Module reuse` section

Thanks for the contribution, however this topic is already covered in the documentation. Perhaps you just missed it

PR #858, opened 2017-07-10
add setup info in installation

vinicius73:


ktsn:
I guess this should be placed in Getting Started section because this is kind of a thing that how to use Vuex rather than installation.

In addition, it would be nice if there is a description that we can use `this.$store` after do that there.

vinicius73:
I agree with you @ktsn. 
I'll move this information and add a few more things.

kiaking:
Closing due to #1468 got merged (which is the update version of this PR).

PR #865, opened 2017-07-16
Add payload property to Payload type

alexjoverm:
The payload property was missing on the Payload type, used for the Mutations/Actions

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -57,7 +57,7 @@ export interface ActionContext<S, R> {
 
 export interface Payload {
   type: string;
-  payload: any;
+  payload?: any;
 }
 
 export interface DispatchOptions {
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -57,6 +57,7 @@ export interface ActionContext<S, R> {
 
 export interface Payload {
   type: string;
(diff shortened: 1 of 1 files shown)


ktsn:
Thanks for the PR. However, the payload can have any properties not only `payload`. It only forces it must have `type` property.

See Object-style commit section: https://vuex.vuejs.org/en/mutations.html

alexjoverm:
Hi @ktsn,

The issue is still not fix. The problem is, for example, writing a plugin, the mutation is not correctly typed (and in this case, it always has a `payload`): 

image

So right now, one has to re-create the type him/herself:

image

That's why I think adding an optional payload parameter (is optional, so no enforcement) solves this issues nicely.

Another option is to extend the type Payload, adding the `payload` property and use it in the subscribe type:

```typescript
  subscribe<P extends Payload>(fn: (mutation: P, state: S) => any): () => void;
```

ktsn:
I see. Then we should declare a new type for the `mutation` since it is different with `Payload` that used on `dispatch`/`commit`.

alexjoverm:
Got it, I'll send a new PR later then. Thanks!

PR #869, opened 2017-07-17
Add MutationPayload type

alexjoverm:
On the subscribe function, the mutation object was only having the `type` property and missing the payload


ktsn:
@alexjoverm Good job 👍 
Can we update the test as well?
https://github.com/vuejs/vuex/blob/dev/types/test/index.ts#L36-L39

alexjoverm:
Sure, there they are!

ktsn:
Great, thank you!

PR #889, opened 2017-07-27
Update structure.md

zanematthew:
Updating to better suite the example; https://github.com/vuejs/vuex/tree/dev/examples/shopping-cart/store and the common approach for using the word `mutation-types.js` rather than `mutations.js`.

### DIFF
--- a/docs/en/structure.md
+++ b/docs/en/structure.md
@@ -23,7 +23,7 @@ For any non-trivial app, we will likely need to leverage modules. Here's an exam
 └── store
     ├── index.js          # where we assemble modules and export the store
     ├── actions.js        # root actions
-    ├── mutations.js      # root mutations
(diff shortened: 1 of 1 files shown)


docs/en/structure.md
```diff
@@ -23,7 +23,7 @@ For any non-trivial app, we will likely need to leverage modules. Here's an exam
 └── store
     ├── index.js          # where we assemble modules and export the store
     ├── actions.js        # root actions
-    ├── mutations.js      # root mutations
```
ktsn: I think `mutations.js` includes mutations itself but not mutation types so we should not remove it.

Also `types.js` would be better since some users want to declare constants for getters and actions, too.

kiaking:
Closing this issue since the example was updated to use module systems now. I think this issue is bit too old to apply to the current structure.

Issue #569, opened 2017-01-07
Update examples with module namespace and add unit tests


alvarengathomas:
First of all, thanks for the hard work on VueJS and Vuex, in the end of january 2017 our company will release a project that uses this frameworks and we are very satisfied with the simplicity and code cleanness of Vue.

We started building the application based on the shopping cart example in October 2016 and we had a hard time trying to separate modules in "namespaces" and to unit test the application, we follow the documentation and use [vue-cli](https://github.com/vuejs/vue-cli) for scaffolding, but the unit tests wasn't working, we had to put a lot of effort and research to finally get things to work properly.

Therefore, my suggestion is to update the examples with this new [namespace feature](https://vuex.vuejs.org/en/modules.html#namespacing) and add unit tests to the examples apps. I believe this will help many other developers to start using VueJS and Vuex.

If you guys consider this to be a important update I would love to do it, I just need a green light from the team!

ktsn:
Hi, thanks for using Vue and Vuex in your project.
I think it is great idea to add the example of namespace feature and unit test. It would be appreciated if you work for it. 🙂

learntoswim:
+1. I'd love to see some advanced testing with vuex actions (including mocking promises).

alvarengathomas:
Thanks for the response guys, I will work on that!

learntoswim:
@alvarengathomas Any progress to share?

alvarengathomas:
Sorry guys, I'm really busy but I'm doing the unit tests in baby steps.

alvarengathomas:
Finally I made the update in shopping cart example to work with module namespace (PR #890)
I will make the unit tests soon!

Eder87rh:
@alvarengathomas do you have the namespace module test? or do you know where can i find an example of namespace module action?

mrmnmly:
I'm also curious how to Unit Test with namespaced module actions

adrienbucci:
Hi everyone,
any news on this topic please ?
@alvarengathomas

mrmnmly:
@adrienbucci what would you like to know? I've already written couple unit tests for my methods (while using namespaces)

adrienbucci:
Thanks @mrmnmly for your message.
I managed to get the namespaces option working, but I can't get my action testing working with this setup.

 ```
expect(jest.fn()).toHaveBeenCalledWith(expected)

Expected mock function to have been called with:
      {"password": "123456", "username": "john"}
    as argument 2, but it was called with
[...]

PR #890, opened 2017-07-27
Update shopping cart example with module namespace

alvarengathomas:
Resulting PR of the discussion in issue #569


examples/shopping-cart/components/Cart.vue
```diff
@@ -14,24 +14,22 @@
...
 <script>
-import { mapGetters } from 'vuex'
+import { mapGetters, mapActions } from 'vuex'
 
 export default {
   computed: {
     ...mapGetters({
-      products: 'cartProducts',
-      checkoutStatus: 'checkoutStatus'
+      products: 'cart/items',
```
ktsn: Maybe we should provide namespace for 1st argument?

```js
...mapGetters('cart', {
  products: 'items',
  checkoutStatus: 'checkoutStatus'
})
```

examples/shopping-cart/components/Cart.vue
```diff
@@ -14,24 +14,22 @@
...
         return total + p.price * p.quantity
       }, 0)
     }
   },
-  methods: {
-    checkout (products) {
-      this.$store.dispatch('checkout', products)
-    }
-  }
+  methods: mapActions({
```
ktsn: We can use namespace as same as the above mapGetters.

```js
mapActions('cart', ['checkout'])
```

examples/shopping-cart/components/ProductList.vue
```diff
@@ -17,13 +17,20 @@ import { mapGetters, mapActions } from 'vuex'
...
-    'addToCart'
-  ]),
+  methods: {
+    ...mapActions({
+      addItemToCart: 'cart/addItem',
+      decreaseProductInventory: 'product/decreaseInventory'
+    }),
+    addToCart(product) {
+      this.addItemToCart(product);
+      this.decreaseProductInventory(product);
```
ktsn: I think we should create `addToCart` action that commits `ADD_ITEM` and `DECREASE_INVENTORY` mutations, and just dispatch it here. Because it shouldn't be separated that adding an item to the cart and decreasing the inventory of added product.

examples/shopping-cart/components/ProductList.vue
```diff
@@ -17,13 +17,20 @@ import { mapGetters, mapActions } from 'vuex'
...
+    }
+  },
   created () {
-    this.$store.dispatch('getAllProducts')
+    this.$store.dispatch('product/getAll')
```
michalsnik: If you already map actions, you can also map this one and simply call `this.getAllProducts()` here. What do you think?

alvarengathomas:
Code updated with the comments suggestions

examples/shopping-cart/store/modules/product/actions.js
```diff
@@ -0,0 +1,12 @@
+import shop from '../../../api/shop'
+import * as types from './types'
+
+export const getAll = ({ commit }) => {
+  shop.getProducts(products => {
+    commit(types.RECEIVE_PRODUCTS, { products })
+  })
+}
+
+export const decreaseInventory = ({ commit }, product) => {
```
ktsn: We seem not to need this action since it's not dispatched in any places?

alvarengathomas: That's true, but what is the best option? Commit mutations from another module like I did OR dispatch actions from another module?

In my opinion the second option (dispatch actions) is better because we don't need to import the types from another module: `import * as productTypes from '../product/types'`

ktsn: Sorry, I totally forgot this PR.

I actually does not use the constant type pattern in Vuex. But I'd say it seems for preventing mismatch between dispatched (committed) type and action (mutation) name caused by typo, refactoring etc, it is ok to use them across different modules.

kiaking:
Currently, cart example already uses the module. Will closing this one 🤝

PR #904, opened 2017-08-12
[zh-hans] use zh-hans instead of zh-cn

lanceli:
Currently the docs of simplified Chinese shown traditional Chinese by gitbook.
Use zh-hans instead of zh-cn to fix it.

Refs: 
- https://github.com/GitbookIO/gitbook/commit/5d1c4d2b9fabedaddaf271f1a5da19aacaaee6c9

-
 https://github.com/GitbookIO/gitbook/blob/b48f8472d35b27d549d7c9a5b61f77f6b78e1908/CHANGES.md#260


Jinjiang:
This will lead the existing zh-cn links broken. I suggest to keep the current links available first and think another way instead.
Thanks

lanceli:
For me it's better to change all existing zh-cn links to zh-hans in the long term, and there is same issue with others docs, e.g. vue-loader

Jinjiang:
I found that the default theme of GItbook don't have `_i18n/zh-cn.json`. So it automatically matches to `_i18n/zh-tw.json`.

My suggestion is create a PR to gitbook `theme-default` to add a `zh-cn.json` file. There are two main reasons: 1. Actually all current content translation is simplified Chinese (except the GitBook theme itself); 2. The `zh-cn` urls has been published.

Thanks.

lanceli:
Up to you, and i do not think they will accept your suggestion base on this: https://github.com/GitbookIO/gitbook/issues/926

PR #905, opened 2017-08-14
[Doc FR] — Core concepts adding and translation

Haeresis:
Hi @posva

We have translated the new « Core Concepts » file.

This PR is ready for merge !


ktsn:
@Haeresis Can we revert `book.json`? Looks like gitbook cannot recognize it since it seems an invalid symbolic link.

Haeresis:
Hi @ktsn 

I do not understand exactly the problem. What can I do to change the problem?

Why my full content was replaced by exactly the same full content for « README.md » and « book.json » ?

ktsn:
@Haeresis 
Hm, I don't know why they are replaced. I just confirm that I cannot serve docs with gitbook locally because of `book.json`.

When I run `yarn docs` (or `npm run docs`), it throws `Error: ENOENT: no such file or directory, stat '/.../vuex/docs/fr/book.json`error.

So I've checked its file type via `ls -l` and it shows that `book.json` is a symbolic link but it points a weird place.

```
lrwxr-xr-x  1 katashin  staff   392 Aug 15 21:39 book.json -> {?  "gitbook": "2.x.x",?  "plugins": ["edit-link", "prism", "-highlight", "github"],?  "pluginsConfig": {?    "edit-link": {?      "base": "https://github.com/vuejs/vuex/tree/dev/docs",?      "label": "Éditer cette page"?    },?    "github": {?      "url": "https://github.com/vuejs/vuex/"?    }?  },?  "links": {?    "sharing": {?      "facebook": false,?      "twitter": false?    }?  }?}?
```

Would it be successfully served in your machine?

Haeresis:
Ok. I delete file and re-inject the previous version. Modifications are not present anymore so, I think it's okay.

ktsn:
It works now. Thanks!

Haeresis:
Thx to you!

PR #914, opened 2017-08-18
fix: move auto installation code into the store constructor

### LINKED ISSUE #731: [vuex] already installed. Vue.use(Vuex) should be called only once.
### Version
2.0.0

### Reproduction link
[https://jsfiddle.net/15ww4vog/8/](https://jsfiddle.net/15ww4vog/8/)

### Steps to reproduce
1. Load jsfiddle page
2. Vue init method is executed.(With CDN, Just VUE and Jquery)
3. Vue init method makes a Ajax call to "https://api.myjson.com/bins/13xym3". and responds a HTML, JS injected to #dinamic-container div.
4. HTML, JS injected also includes a Vue instance.
5. creating conflicts "[vuex] already installed. Vue.use(Vuex) should be called only once."

### What is expected?
1. Load page, instantiate VUE.
2. When a page is injected and includes a VUE instance. load with no conflicts.

1. Load page, instantiate VUE.
2. When page is injected and includes a VUE instance. a conflict is raised and show the "[vuex] already installed. Vue.use(Vuex) should be called only once." console error.

---
My parent view is loading Vue instance.
and I'm injecting a Child view that also contains its own Vue instance.

ktsn:
While auto-calling `Vue.use(Vuex)` is convenient for users who use Vue/Vuex via `<script>`, there are a case that users want to avoid the behavior when `window.Vue` is exists (See #731).

Currently, Vuex is immediately installed into `window.Vue` when Vuex is imported. So there are no way to install Vuex into `import`ed Vue.

To solve this, I'd move it in `Vuex.Store`'s constructor. Then the auto-installation is delayed until users create a store and the users can install Vuex into their own Vue constructor before doing it.

I'm not sure how to write test code for this but it looks like working well.

fix #731

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -7,6 +7,13 @@ let Vue // bind on install
 
 export class Store {
   constructor (options = {}) {
+    // Auto install if it is not done yet and `window` has `Vue`.
+    // To allow users to avoid auto-installation in some cases,
+    // this code should be placed here. See #731
+    if (!Vue && typeof window !== 'undefined' && window.Vue) {
+      install(window.Vue)
+    }
+
     if (process.env.NODE_ENV !== 'production') {
       assert(Vue, `must call Vue.use(Vuex) before creating a store instance.`)
       assert(typeof Promise !== 'undefined', `vuex requires a Promise polyfill in this browser.`)
@@ -462,9 +469,4 @@ export function install (_Vue) {
   }
   Vue = _Vue
   applyMixin(Vue)
-}
-
-// auto install in dist mode
-if (typeof window !== 'undefined' && window.Vue) {
-  install(window.Vue)
 }


yyx990803:
Since the constructor can be called multiple times, we need to make sure `install` is called only once.

ktsn:
I thought existence check of inner `Vue` is enough because when `install` is called, the inner `Vue` will be assigned. Then `install` will be skipped on the next time.
Am I missing any other cases?

yyx990803:
Ah right, nevermind!

ianrussel:
Ihave same problem 
   `[vuex] already installed. Vue.use(Vuex) should be called only once. `

When using window.Vue

I fixed it by 

     `window.Vue = require('vue');
      window.Vuex = require('vuex');
      Vue.use(Vuex); `

PR #919, opened 2017-08-27
added gitbook install in docs script

steven-ferrer:
Installing plugin dependencies before doing `gitbook serve`

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "test:e2e": "node test/e2e/runner.js",
     "test:types": "tsc -p types/test",
     "release": "bash build/release.sh",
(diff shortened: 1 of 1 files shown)


ktsn:
Hmm, I personally prefer not include install script in it because it does not need to execute in every time.

steven-ferrer:
hmm, it wouldn't really install the dependencies every time if they already exists. anyway feel free to close this :)

ktsn:
Right, but it takes a few time even if they are already installed. It's same reason we do not include `npm install` in build command 🙂 
Anyway, thanks for your PR!

PR #920, opened 2017-08-27
Update api.md: take createNamespacedHelpers out

tsaarikivi:
createNamespacedHelpers seems to not belong to 'vuex' package


ktsn:
Thanks for the PR. But it actually is in Vuex. It just is not released yet.

PR #922, opened 2017-08-28
fix(typings): watch() returns an unwatch function

FOODy:
Corrected the return type of the `Store#watch` method in the typescript typings.


ktsn:
Thanks!

PR #924, opened 2017-08-30
feat: allow to passing functions in mapActions/mapMutations (fix #750)

### LINKED ISSUE #750: Allow function for mapMutations/mapActions like mapState
in function normalizeNamespace

make payload to string is so pain;

can we add the typeOf(val) to judege this type?

my English is poor, please forgive me

I want to use mapMutations set payload mutations,like

Vuex.mapMutations({
                increame: {
                    type: 'increame',
                    addNum: 10
                }
            })

ktsn:
This is a small improvement for `mapActions` and `mapMutations` helpers. It allows users to pass functions in the helpers so that reducing some redundant method definitions. The feature is like a counterpart of passing function in `mapState`.

For example:

```js
const vm = new Vue({
  store,
  template: `<input @input="onInput">`
  methods: mapActions({
    onInput (dispatch, event) {
      // First argument is as same as this.$store.dispatch
      dispatch('inputText', {
        text: event.value
      })
    }
  })
})
```

The above code is equivalent with the below (Note that `inputText` is a bit redundant since it is just an bridge method between `onInput` and the action):

```js
const vm = new Vue({
  store,
  template: `<input @input="onInput">`
  methods: {
    ...mapActions(['inputText']),
    onInput (event) {
      this.inputText({
        text: event.value
      })
    }
  })
})
```

In addition, this syntax is more effective when we map namespaced modules.

```js
methods: mapActions('some/module', {
  onInput (dispatch, event) {
    // `dispatch` is already namespaced here.
    // So the below line will call `some/module/inputText` action.
    dispatch('inputText', {
      text: event.value
    })
  }
})
```

I did not implement this syntax for `mapGetters` because we already can do the same thing with `mapState`.

Close #750

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -25,12 +25,18 @@ export const mapState = normalizeNamespace((namespace, states) => {
 export const mapMutations = normalizeNamespace((namespace, mutations) => {
   const res = {}
   normalizeMap(mutations).forEach(({ key, val }) => {
-    val = namespace + val
     res[key] = function mappedMutation (...args) {
-      if (namespace && !getModuleByNamespace(this.$store, 'mapMutations', namespace)) {
-        return
+      let commit = this.$store.commit
+      if (namespace) {
+        const module = getModuleByNamespace(this.$store, 'mapMutations', namespace)
+        if (!module) {
+          return
+        }
+        commit = module.context.commit
       }
-      return this.$store.commit.apply(this.$store, [val].concat(args))
+      return typeof val === 'function'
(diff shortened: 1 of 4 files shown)


rayax86:
Hi @ktsn, could you please elaborate on 

> I did not implement this syntax for `mapGetters` because we already can do the same thing with `mapState`.

?

rayax86:
Nevermind, I now see getters are passed to `mapState` as a param.

Would be nice if the documentation mentions this though.

PR #927, opened 2017-08-31
fix: initialize root state as an empty object if state function returns no value

n7olkachev:
Just a moment ago I had hard time debugging this code:
```js
const state = () => {}
...
new Vuex.Store({
    state,
    modules,
    ...
})
```
If state is `undefined` and you use modules, Vuex can't create store. This small change will make it easier to find this error.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -24,7 +24,7 @@ export class Store {
     if (typeof state === 'function') {
       state = state()
     }
-    
+
     if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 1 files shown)


ktsn:
Thank you for your contribution. However, I think providing an empty object as a default value is more appropriate here because modules' state actually behaves so.
https://github.com/vuejs/vuex/blob/dev/src/module/module.js#L9

n7olkachev:
Is it ok now?

Issue #908, opened 2017-08-15
Development mode: Ability to log actions/mutations within Vue devtools


b44rd:
### What problem does this feature solve?
Referencing to https://github.com/vuejs/vue-devtools/issues/179

A pretty sweet development feature would be to add another tab inside Vue devtools that track the flow of actions and mutations. The suggestion is to allow a function such as devtools.log inside both store.dispatch and store.commit that holds the passed params. I´ve been using a custom library to do this for a while (b44rd/jsbug), and it sure provides another level of debuggability. The problem though, is that I currently need to add a line of code inside each action/mutation to achieve this overview. 

For instance, when dispatching an action, this might look like this inside a new tab inside Vue devtools: 
```
▶ Action: [actionName] [timestamp]
```

When clicked, this can be expanded, such as
```
▼ Action: [actionName] [timestamp]
  {
    points: 1
  }
```

The result list then visible in Vue devtools might then look like this. Note: Making actions print in a different color than mutations makes it really easy to follow the flow
```
▶ Action: "init" 10:00:002
▶ Mutation: "INIT" 10:00:004
▶ Action: "jump" 10:01:026
▶ Mutation: "JUMP" 10:01:031
▼ Action: "collectPoint" 10:02:005
  {
    points: 1
  }
▼ Mutation: "SET_STATUS" 10:02:015
  {
    points: 1,
    status: 'hero'
  }
```

### What does the proposed API look like?
I believe it can be an alternative to extend Vue.config to enable/disable this feature. Since Store.prototype.commit is already logging to the mutation panel, I believe Store.prototype.dispatch can also be modified to do the same, so we will basically be able to call something similar to these two when dispatching actions or committing mutations
```
devtools.logAction('actionName', payload)
devtools.logMutation('mutationName', payload)
```

posva:
Isn't this achieved with https://vuex.vuejs.org/en/plugins.html#built-in-logger-plugin?

b44rd:
@posva Thanks! Will have a look at this one. I´ve not studied the code enough to realize what´s fully happening in order for vue-devtools to pick this up. But isn´t this plugin mainly focusing on mutations? The overall idea of this feature request is to achieve a log that also tracks actions, so the action -> mutation flow somehow becomes debuggable inside a theoretical devtools panel.

posva:
Actions are not logged because they don't affect the state, and mutations can happen asynchronously inside of actions

b44rd:
[...]

PR #929, opened 2017-09-04
Proposal: Global action enhancers - WIP

deini:
Proposal to bring Global action enhancers.

I wanted a way to log all actions and their payload when dispatched. Following the discussion in https://github.com/vuejs/vuex/issues/908 seems like there is no way right now to do cleanly.

The proposal is to bring global action enhancers that work just like redux's middleware.

Using the count example and some inline actionEnhancers:

```js
export default new Vuex.Store({
  state,
  getters,
  actions,
  mutations,
  actionEnhancers: [
    context => next => (action, payload) => {
      console.log('Hello from first enhancer:', context.state.count)
      return next(action, payload)
    },
    context => next => (action, payload) => {
      console.log('Hello from second enhancer:', context.state.count)
      return next(action, payload)
    }
  ]
})
```

If we can get something like this, then adding an actions tab in devtools would be pretty straight forward.

TODO:
- Feedback
- Docs
- Tests

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -1,7 +1,7 @@
 import applyMixin from './mixin'
 import devtoolPlugin from './plugins/devtool'
 import ModuleCollection from './module/module-collection'
-import { forEachValue, isObject, isPromise, assert } from './util'
+import { compose, forEachValue, isObject, isPromise, assert } from './util'
 
 let Vue // bind on install
 
@@ -21,6 +21,7 @@ export class Store {
     }
 
     const {
+      actionEnhancers = [],
       plugins = [],
       strict = false
     } = options
@@ -42,9 +43,14 @@ export class Store {
     this._subscribers = []
     this._watcherVM = new Vue()
 
-    // bind commit and dispatch to self
     const store = this
-    const { dispatch, commit } = this
+    const { commit } = this
+    let { dispatch } = this
+
+    const actionEnhancerChain = getActionEnhancerChain(store, state, actionEnhancers)
+    dispatch = compose(...actionEnhancerChain)(dispatch.bind(store))
+
+    // bind commit and dispatch to self
     this.dispatch = function boundDispatch (type, payload) {
       return dispatch.call(store, type, payload)
     }
@@ -436,6 +442,15 @@ function enableStrictMode (store) {
       assert(store._committing, `Do not mutate vuex store state outside mutation handlers.`)
     }
   }, { deep: true, sync: true })
+}
+
+function getActionEnhancerChain (store, state, actionEnhancers) {
+  return actionEnhancers.map(actionEnhancer => actionEnhancer({
+    dispatch: store.dispatch,
+    commit: store.commit,
+    getters: store.getters,
+    state
+  }))
 }
 
 function getNestedState (state, path) {
--- a/src/util.js
+++ b/src/util.js
@@ -64,3 +64,26 @@ export function isPromise (val) {
 export function assert (condition, msg) {
   if (!condition) throw new Error(`[vuex] ${msg}`)
 }
+
+/**
+ * Taken from Redux repo
+ * Composes single-argument functions from right to left. The rightmost
+ * function can take multiple arguments as it provides the signature for
+ * the resulting composite function.
+ *
+ * @param {...Function} funcs The functions to compose.
+ * @returns {Function} A function obtained by composing the argument functions
+ * from right to left. For example, compose(f, g, h) is identical to doing
+ * (...args) => f(g(h(...args))).
+ */
+export function compose (...funcs) {
+  if (funcs.length === 0) {
+    return arg => arg
+  }
+
+  if (funcs.length === 1) {
+    return funcs[0]
+  }
+
+  return funcs.reduce((a, b) => (...args) => a(b(...args)))
+}


ktsn:
Hi @deini. Thanks for your contribution!
I just read #908 and I think the action tab in devtools is very nice idea 👍 

One thing I'm concerning is that global action enhancer seems a bit overkill to me. It potentially allows users to mutate all the dispatched actions and can complicates the actions flow. 

As we just want to listen the `dispatch`, wouldn't it enough to add `subscribe`-like method but for action? It also have advantage that it can be utilized in some plugins.

```js
function actionLoggerPlugin(store) {
  // Not sure how we should call this method...
  store.subscribeAction(({ type, payload }, state) => {
    console.log(`Before dispatching "${type}" with ${JSON.stringify(payload)}`
  })
}

  // ...
  plugins: [actionLoggerPlugin]
})
```

deini:
@ktsn I agree, global action enhancers are overkill for #908, however it could open up for more middleware use cases. It indeed complicates things more and you are right that a user can mutate / discard actions.

I like your proposal, as its all we need for #908. Not sure if you guys had any other use cases for GAE, but if not, feel free to close and I can give `subscribeAction` a shot.

yyx990803:
`subscribeAction` sounds good to me 👍

LinusBorg:
I'm currently playing around with the ideas from `redux-first-router` and try to come up with a Vue&vuex solution.

This pattern uses actions as triggers for route navigation, so these proposed actionEnhancers would actually be perfect to implement navigation guards, redirects etc. Currently I monkeypatch `dispatch()` to be able to do that.

a `subscribeAction` as suggested by @ktsn would not work here, as I would not have any way of cancelling the action.

ktsn:
@LinusBorg Thanks for the idea! I didn't know `redux-first-router` but it looks nice concept 👍 . I agree that it would be also useful in Vue & Vuex.

I'm thinking how we can implement that concept in Vuex and it could be solved without enhancer/middleware feature. PoC is https://github.com/ktsn/vuex-first-router
The basic idea is to separate the actions with *route change request* and *route changed event*.

For example, if we have `foo` route and want to move to it, we dispatch `foo` action (route change request). Then the router changes the url and dispatches another action - `applied:foo` action (route changed event). 
If `foo` actually redirects to `bar`, we dispatch `foo` action but `applied:bar` will be dispatched after that.

I'm not sure if we need to use the same action name to achieve this concept. But to my understanding, the implementation would be simpler by separating the actions because we don't need to branch the further processing where the action comes from (changing url on address bar or directly dispatching it).

Maybe I'm a bit harsh with this topic 😅 . But I'd say the action already can do various things out of the box, not sure the trade off between the benefit and additional complexity.

lmiller1990:
I initially asked on [stackoverflow](https://stackoverflow.com/questions/47066028/call-mutation-before-every-action-vuex), but I want a way to do a generic something before actions - basically, similar to this feature, or an action logger.

Any update on this PR?

jannesiera:
Any progress on this?

I'm currently looking for a way to :

1) automatically call mutations if an action of the same name doesn't exsist (so my calling code always dispatches, never commits).
2) prevent commits from being accepted with a decoupled guard function.

I got this to work to overwrite vuex's dispatch and commit methods. However, it would be nice if I could achieve both in a supported and standardized way.

jannesiera:
Bump.

Ran into another use case where this would be very useful. While creating a small ORM to handle mutations of the state with all internal relations in a standard way, it would be very handy if I could pass the state to the ORM manager before every mutation.

Currently, I'd have to either (1) call the function at the beginning of every mutation manually or (2) overwrite the dispatch/commit methods again.

Option 1 is an unacceptable amount of boilerplate that I couldn't afford in any professional project.
Option 2 will make it virtually impossible to share my solution with the wider Vuex community.

lmiller1990:
I posted earlier asking for a way to subscribe to actions which was indeed added in 2.5 - `subscribeAction`.

@jannesiera couldn't you write a plugin that just hooks into `store.subscribeAction` to achieve the goal you want?

```js
export default function() {
  return store => {
    store.subscribeAction(action, state) => {
      // do stuff
    }
  }
}
```

zgayjjf:
@lmiller1990 
subscribeAction can only executed after action executed.
What @jannesiera want was something executed before action, which I think would be great if there is someway.

ta4:
How we can cancel action or modify payload inside store.subscribeAction hook?

Gvade:
@jannesiera 
In order to write some sort of middleware for your actions or mutations you can use the standard [ES2015 Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy). For instance:

```
...
const actions = {
    close(context, payload) {
       //your action
    }
}

actions['close'] = new Proxy(actions.close, {
    apply: function(target, thisArg, argumentsList) {
        const [context, payload] = argumentsList;
        //your middleware code
        return target.apply(thisArg, argumentsList);
    }
});
...
```

But if I understand correctly, the middleware logic can be put just at the beginning of an action:
```
...
const actions = {
    close(context, payload) {
        //middleware code
        
        //your action
    }
}
```

kiaking:
Currently, Vuex has `subscribeAction` with `before` option where you can do something before an actions gets executed. I think this PR could be closed due to its over age.

I'll keep this issue open for now to gather feedbacks if there's any.

kiaking:
Closing due to inactivity. I think this can be done by `subscribeAction` 👍

PR #931, opened 2017-09-05
Update package.json

dear-lizhihua:
fix `Error: Couldn't locate plugins "edit-link, prism, github, theme-vuejs", Run 'gitbook install' to install plugins from registry.`


chrisvfritz:
Since we use Gitbook for so many repos, I wonder if it'd actually be good to extract some common config/build process out into a package. Not volunteering, but putting the idea out there... 😅

chrisvfritz:
Actually, I wonder if this is a job that calls for @atomaka. 😉 

Server Superhero

dear-lizhihua:
I encountered a mistake when I started the project. I looked at the configuration of vue-router, it's OK. So I created this issue.

yyx990803:
I made the same fix when I was updating the docs, but thanks for the PR!

PR #941, opened 2017-09-11
Namespaced modules: root actions

### LINKED ISSUE #770: Proposal: root actions for namespaced module
I have app that contain dynamic views (kinda of mini app). Each view have its own module.
Also , in my app have a global filters of the data each view need to show.

Now , every time I change the global filters ,i need to make sure my dynamic view is refresh its own data.
Until now every module implement its own "refresh" action.

Since every dynamic view is like mini-app here , we decide to make them namespaced.
Now when "refresh" is dispatched my module didn't get the "global refresh event".

Simply what I want is to enable the namespaced module to register the some actions under the root namespace. Its enable namespaced modules to participate in root events while changing the namespaced module. 
The root actions get the same context of the regular ones.

It's  easy to implement in vuex , the question if this is necessary

P.S

maybe do the same for mutations

Root propertry :
```javascript
  modules: {
    files: {
      namespaced: true,

      // module assets
      state: { ... }, // module state is already nested and not affected by namespace option
      getters: {
        bigFiles () { ... } // -> getters['files/bigFiles']
      },
      actions: {
        delete (namespacedContext) { ... }, // -> dispatch('files/delete')
        refresh: {
            root: true,
            handler (sameNamespacedContext) { ... } // -> dispatch('refresh')
        }
      },
      mutations: {
        delete () { ... } // -> commit('files/delete')
      }

    }
  }
})
```

or new "rootActions" :
```javascript
  modules: {
    files: {
      namespaced: true,

      // module assets
      state: { ... }, // module state is already nested and not affected by namespace option
      actions: {
        delete (namespaceContext) { ... } // -> dispatch('files/delete')
      },
      rootActions: {
        refresh (sameNamespaceContext) { ... } // -> dispatch('refresh')
      },
      mutations: {
        delete () { ... } // -> commit('files/delete')
      }

    }
  }
})
```

Akryum:
Fix #770 

The actions can now be defined with the following notation:

```javascript
{
  root: true,
  handler (context, payload) { ... }
}
```

- [x] Tests
- [x] Type def


src/store.js
```diff
@@ -287,8 +287,19 @@ function installModule (store, rootState, path, module, hot) {
   })
 
   module.forEachAction((action, key) => {
-    const namespacedType = namespace + key
-    registerAction(store, namespacedType, action, local)
+    let finalType, finalAction
```
ktsn: Maybe we can just write like following?

```js
const type = action.root ? key : namespace + key
const handler = action.handler || action
registerAction(store, type, handler, local)
```

src/module/module-collection.js
```diff
@@ -84,25 +84,43 @@ function update (path, targetModule, newModule) {
...
+  expected: 'function'
+}
+
+const objectAssert = {
+  assert: value => typeof value === 'function' ||
+    (typeof value === 'object' && typeof value.handler === 'function'),
+  expected: 'function or object with "handler" function'
+}
+
+const assertTypes = {
```
ktsn: This is nice assertion approach 👍

PR #949, opened 2017-09-13
state.cart is undefined. update to state.added

connor11528:
update example code for shopping cart


ktsn:
It should be `state.cart.added` since the state is in `cart` module...
Doesn't it work in your machine?

connor11528:
ohhhh okay. I was getting some other errors that were unrelated to this. My bad!

PR #960, opened 2017-09-19
feat: subscribeAction

deini:
Allows to listen dispatched actions

After feedback from https://github.com/vuejs/vuex/pull/929 so we can implement #908

API is as suggested by @ktsn 
```js
function actionLoggerPlugin(store) {
  store.subscribeAction(({ type, payload }, state) => {
    console.log(`Before dispatching "${type}" with ${JSON.stringify(payload)}`
  })
}

  // ...
  plugins: [actionLoggerPlugin]
})
```

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -35,6 +35,7 @@ export class Store {
     // store internal state
     this._committing = false
     this._actions = Object.create(null)
+    this._actionSubscribers = []
     this._mutations = Object.create(null)
     this._wrappedGetters = Object.create(null)
     this._modules = new ModuleCollection(options)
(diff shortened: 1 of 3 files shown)


deini:
Made some changes to the built-in Logger plugin to play a bit more with the idea.

image

Not sure if I should include those changes in this PR since I'm not sure if this is something you guys want to support out of the box or would prefer going with  a 3rd party `actionLoggerPlugin`. 

Changes are [here](https://github.com/deini/vuex/compare/subscribeAction...deini:logger-plugin-actions?expand=1) if anyone is interested.

PR #965, opened 2017-09-22
Update translation. Remove incorrect using of the term "pure function"

sergey-bastrakov:
The term "pure function" ("чистые функции" in russian) used incorrectly in the translation. A mutation can not be a pure function, because it changes the state of the passed argument. This term should be translated as "просто функции" or "функции" ("just function" in english).


sergey-bastrakov:
https://github.com/translation-gang/ru.vuex/pull/35

PR #973, opened 2017-10-02
Update modules.md

ahalimkara:


ktsn:
Thanks for the PR.
The current comment is however correct because the myPage module does not have `namespaced: true` option.

ahalimkara:
Thank you for explanation, i just missed that part.

PR #987, opened 2017-10-14
Proposal: createLogger can optionally log actions

deini:
## Proposal
Allow `createLogger` to optionally log actions, added a couple of options
```js
{
  actionFilter,
  actionTransformer,
  logActions,
  logMutations
}
```

~With some sane default to keep current functionality by default, this would be an opt-in option.~ Enabled by default as requested.

When enabled you get something like:

It just logs the action type and payload, since we cant tie actions to mutations, which I think its ok.

Note: Also added the `devtoolHook` for actions so we can hook it up later with devtools.

PS: I'm about to do a small plugin in case this proposal doesn't get approved and will update this PR with the url just in case someone is looking for this functionality. It would be awesome if we could have it built-in tho.

### DIFF
--- a/src/plugins/logger.js
+++ b/src/plugins/logger.js
@@ -7,47 +7,81 @@ export default function createLogger ({
   filter = (mutation, stateBefore, stateAfter) => true,
   transformer = state => state,
   mutationTransformer = mut => mut,
+  actionFilter = (action, state) => true,
+  actionTransformer = act => act,
+  logActions = false,
+  logMutations = true,
   logger = console
 } = {}) {
   return store => {
     let prevState = deepCopy(store.state)
 
-    store.subscribe((mutation, state) => {
-      if (typeof logger === 'undefined') {
-        return
-      }
-      const nextState = deepCopy(state)
+    if (typeof logger === 'undefined') {
+      return
+    }
 
-      if (filter(mutation, prevState, nextState)) {
-        const time = new Date()
-        const formattedTime = ` @ ${pad(time.getHours(), 2)}:${pad(time.getMinutes(), 2)}:${pad(time.getSeconds(), 2)}.${pad(time.getMilliseconds(), 3)}`
-        const formattedMutation = mutationTransformer(mutation)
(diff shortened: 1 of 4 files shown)


src/plugins/logger.js
```diff
@@ -7,49 +7,83 @@ export default function createLogger ({
...
+
+    if (logActions) {
+      store.subscribeAction((action, state) => {
+        const currentState = deepCopy(state)
 
-        try {
-          logger.groupEnd()
-        } catch (e) {
-          logger.log('—— log end ——')
+        if (actionFilter(action, currentState)) {
```
ktsn: I guess we don't need to deep copy the state here since not retain `prevState` for actions.

src/plugins/logger.js
```diff
@@ -7,49 +7,83 @@ export default function createLogger ({
   filter = (mutation, stateBefore, stateAfter) => true,
   transformer = state => state,
   mutationTransformer = mut => mut,
+  actionFilter = (action, state) => true,
+  actionTransformer = act => act,
+  logActions = false,
```
ktsn: Maybe it's ok to also enable `logActions` in default?

src/plugins/logger.js
deini: Good call 💯

deini:
@ktsn Thanks for the review, addressed your comments, let me know if you want me to squash my commits.

jeremyjh:
What is the status of this, will it be merged?

GufNZ:
Dangit - should have looked here first - I just wrote the same thing `:-p`

jandob:
Almost 2 years later: 
Again, will it be merged?

kiaking:
Sorry for the late response on this one, though it might be nice idea to have this...? It's pretty old so it might not be possible but @deini do you think you can rebase this one to the latest `dev` branch?

deini:
@kiaking rebased

kiaking:
Thanks a lot! Checked locally as well and it's working 👍

PR #988, opened 2017-10-14
Update docs/ja/forms.md

38elements:


docs/ja/forms.md
```diff
@@ -21,7 +21,7 @@ computed: {
   })
 },
 methods: {
-  updateMessage: ({ dispatch }, e) => {
+  updateMessage: (e) => {
```
ktsn: Thanks for correcting this!
Would you mind also change this to a shothand method style definition? The arrow function is still invalid here because we cannot use `this` as a component instance in it.

38elements: Thank you for reviewing.
I fixed it.

PR #1000, opened 2017-10-17
Fix export default in index.d.ts

### LINKED ISSUE #999: Typescript 2.6 errors on `export default`
### Version
3.0.0

### Reproduction link
[https://github.com/Microsoft/TypeScript/pull/18444](https://github.com/Microsoft/TypeScript/pull/18444)

### Steps to reproduce
* git clone https://github.com/vuejs/vuex.git
* cd vuex
* npm install
* npm install typescript@next
* cd types
* tsc -p test/tsconfig.json

### What is expected?
No error.

Error: The expression of an export assignment must be an identifier or qualified name in an ambient context.

---
This error is new in Typescript 2.6. It disallows the export of expressions, including object literals like `{ Store, install }`. We found this error when testing popular packages with Typescript 2.6 RC.

sandersn:
The default export in index.d.ts breaks in Typescript 2.6 with the error "The expression of an export assignment must be an identifier or qualified name in an ambient context."

This commit also adds a small test that uses the default export, since no test existed previously.

Fixes #999


types/index.d.ts
```diff
@@ -121,7 +121,8 @@ export interface ModuleTree<R> {
   [key: string]: Module<any, R>;
 }
 
-export default {
-  Store,
-  install
-};
+export const _default: {
```
ktsn: Why `_default` is also exported?

sandersn: Oops! My mistake. I'll remove it.

ktsn:
Thanks!

skray:
Hi @ktsn do you have an idea of when this will make it into a release? I just ran into this while upgrading to TS 2.6 and I'm not sure if there's any resolution except for pinning to TS 2.5 until this fix is available.

lishewen:
Please publish this change.
vuex 3.0.0 no support typescript 2.6

ktsn:
v3.0.1 has been published.

PR #1005, opened 2017-10-18
update docs of state & mutation & action

DuLinRain:
1, point out an unmentioned way of mapMutations helper mapping 
2, point out an unmentioned way of mapActions helper mapping 
3. point out that getters will be exposed as second parameter when maping States as normal function
4. point out that mapMutations can also receive a module name as the first argument so as to commit 
    mutattion to a module with namespace.
5. point out that mapActions can also receive a module name as the first argument so as to commit 
    mutattion to a module with namespace.
6. point out that mapState can also receive a module name as the first argument so as to commit 
    mutattion to a module with namespace.

### DIFF
--- a/docs/en/actions.md
+++ b/docs/en/actions.md
@@ -129,7 +129,7 @@ export default {
   methods: {
     ...mapActions({
       addAlias: function (dispatch) {
-        dispatch('increment') // map `this.addAlias(amount)` to `this.$store.dispatch('increment', amount)`
+        dispatch('increment') // map `this.addAlias()` to `this.$store.dispatch('increment')`
       }
     })
   }
@@ -155,7 +155,7 @@ export default {
     }),
     ...mapActions('moduleName', {
       addAlias: function (dispatch) {
-        dispatch('increment') // map `this.addAlias(amount)` to `this.$store.dispatch('increment', amount)`
+        dispatch('increment') // map `this.addAlias()` to `this.$store.dispatch('increment')`
       }
     })
   }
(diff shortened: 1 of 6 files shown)


Jinjiang:
如果我没有弄错的话，这些内容是英文文档版本里没有的。我们目前对于其它语言版本的文档都是翻译自英文版。如果你觉得上述改动是值得在所有语言版本中呈现的，那么建议优先提供英文版的修改作为讨论。
谢谢：）
英文文档地址：https://github.com/vuejs/vuex/tree/dev/docs/en

akanshgulati:
@DuLinRain  Really good additions as I tried to look at few diffs. But, when I compare it with present version, I believe that mentioning a `getter` function in `state` could confuse user who is reading it for the first time. I am saying this because when I tried to relate the additions from current page of `state` I could only understood about the example including `getters` because I read about it first. 
@Jinjiang Please review this.

Jinjiang:
@akanshgulati yes I agree with you. Maybe put all helpers out of "core concepts" with a separated article is better IMO. So people will know all state, getter, module first, and then understand what the helpers do and the API arguments.

kiaking:
I like the idea of separating helpers to its own section as mentioned by @Jinjiang.

@DuLinRain Sorry for the late action but do you think you can revise the PR to be based from the latest `dev` branch...? Or maybe it would be faster to recreate a new PR from scratch due to its age 😓

akanshgulati:
Quite a time I must say for reviewing it.

kiaking:
I guess so. I'll wait for about week and if there's nothing happening, I'll close this one 👍

PR #1031, opened 2017-10-30
wording did not make sense.

playpianolikewoah:
> Actions don't

Did not make sense.

### DIFF
--- a/docs/en/actions.md
+++ b/docs/en/actions.md
@@ -45,7 +45,7 @@ Actions are triggered with the `store.dispatch` method:
 store.dispatch('increment')
 ```
 
-This may look dumb at first sight: if we want to increment the count, why don't we just call `store.commit('increment')` directly? Well, remember that **mutations must be synchronous**? Actions don't. We can perform **asynchronous** operations inside an action:
+This may look dumb at first sight: if we want to increment the count, why don't we just call `store.commit('increment')` directly? Remember that **mutations must be synchronous**? Actions need not be. We can perform **asynchronous** operations inside an action:
 
 ``` js
 actions: {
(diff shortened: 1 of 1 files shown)


posva:
Thanks for your interest in helping the docs, but that is valid English 🙂

_The sky is blue, the sun isn't_

playpianolikewoah:
@posva, I understand what the intention is, however, it is not valid English. When I first read it, it confused me.

I the docs said: 
> mutations have to be synchronous. Actions don't

It would make sense. 
Also the abruptness of the second sentence will confuse the reader, as it did me. 
Also the example you gave when you closed this PR is not the same structure. 

The main reason it doesn't make sense is if you complete the sentence. 

> Well, remember that **mutations must be synchronous**? Actions don't must be synchronous. 

Changing the "must" to "have" and completing the sentence you get: 

> Well, remember that **mutations have to be synchronous**? Actions don't have to be. 

Or as I suggested also works without changing the first sentence. 

Either way the second sentence is unnecessarily abrupt.

posva:
You mean it could be

> Well, remember that mutations must be synchronous? Actions mustn't
?

I personally think adding  _be synchrounous_ at the end makes the sentence heavy but I'm not native. WDYT @chrisvfritz @yyx990803

playpianolikewoah:
@posva That is an option. However, I offered other options that sound better, IMO. 
I personally think: 
> mutations have to be synchronous. Actions don't. 

Is the easiest fix.

chrisvfritz:
@posva I might be late, but I like that as well. 🙂 

> mutations have to be synchronous. Actions don't.

PR #1042, opened 2017-11-03
simply removed an extra "do"

csheets:


ktsn:
Thank you for interested in improving the docs.
However it is valid English. It is used for emphasizing the verb.

PR #1047, opened 2017-11-06
docs: update watch() description

marszall87:
Description of the `watch()` function is not accurate, [getter receives two arguments](https://github.com/vuejs/vuex/blob/dev/src/store.js#L150).


ktsn:
Good catch!

Issue #863, opened 2017-07-13
Allow function to be passed as namespace argument for mapState, mapGetters, mapMutations & mapActions


aspruds:
### What problem does this feature solve?
Right now I can set namespace when calling mapState. However, mapState only accepts string as a namespace. Unfortunately, I can not use Vue props values as these are only available after the component has been created. It means that I need to hardcode Vuex namespace and it is impossible to reuse the component.

### What does the proposed API look like?
I would like the following functionality:
```javascript
props: {
    namespace: {
        type: String,
        default: 'searchForm'
    }
},
computed: {
     // namespace in a function
  ...mapState(function() { return this.namespace }, ['username', 'password'])
},
```
To implement this, one would need to change mapState implementation. normalizeNamespace function would need to be called later (inside mappedState). Since mappedState is called only after the component has been initialized, it would be possible to resolve .namespace prop by calling the passed function.

I could provide a small code sample with my implementation (rewrite of current mapState) if that would help in any way.

ktsn:
ref #748

LinusBorg:
You could use `mapState` et al. in `beforeCreate()` to dynamically extend `this.$options.computed

aspruds:
Another workarround which I am considering is to use vuex-connect.

ktsn:
I think this is covered by passing a function to mapState/mapActions/mapMutations (related #924). 
After releasing #924, we can dynamically change namespace of all store assets in mapXXX helpers.

```js
{
  props: ['namespace'],

  computed: mapState({
    state (state) {
      return state[this.namespace]
    },
    someGetter (state, getters) {
      return getters[this.namespace + '/someGetter']
    }
  }),

  methods: {
    ...mapActions({
      someAction (dispatch, payload) {
        return dispatch(this.namespace + '/someAction', payload)
      }
    }),
    ...mapMutations({
      someMutation (commit, payload) {
        return commit(this.namespace + '/someMutation', payload)
      })
    })
  }
}
```

men62:
Will the dispatch return promise? If yes, personally, is it good practice to assign the data from that?

johnsoncheg:
@ktsn I'm surprised that when I use arrow function `state => state[this.namespace]`,  some errors about the this binding happen, I feel confused

rATRIJS:
@ktsn it won't work neatly for deeper namespaces - e.g. `level1/level2/level3` - as it would try to access it incorrectly:

```js
mapState({
[...]

PR #1048, opened 2017-11-06
Added dynamic namespacing to mapState, mapGetters, mapActions

Stumblor:
Based on this issue [https://github.com/vuejs/vuex/issues/863](https://github.com/vuejs/vuex/issues/863), I was also struggling to apply a dynamic namespace to a component that could be used anywhere in the component tree. To support the namespace being applied to a component via the props, I added functionality for the namespace to be supplied as a function.

Example usage:
```javascript
    var namespace_def = "settings/someModule/";

    module.exports = {
        props: [ "namespace" ],
        computed: mapState((self) => self.namespace || namespace_def, [
            'options', 
            'items'
        ]),
        methods: mapActions((self) => self.namespace || namespace_def, [
            'SomeAction', 
            'AnotherAction'
        ])
     };
```

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -5,6 +5,8 @@ export const mapState = normalizeNamespace((namespace, states) => {
       let state = this.$store.state
       let getters = this.$store.getters
       if (namespace) {
+        if (typeof namespace === 'function') namespace = namespace.apply(this, [this])
+        if (namespace.charAt(namespace.length - 1) !== '/') namespace += '/'
         const module = getModuleByNamespace(this.$store, 'mapState', namespace)
         if (!module) {
           return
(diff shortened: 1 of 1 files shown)


ktsn:
IMO, it would be better to use function-style mapper described on [this comment](https://github.com/vuejs/vuex/issues/863#issuecomment-329510765) instead of introducing dynamic namespace.
Because the latter declares the component computed/methods dynamically. It would make harder to  type check statically and grasp the component behavior.

Stumblor:
That's true, but it's at the expense of a more verbose declaration. I can see this becoming more of a problem will large numbers of methods / properties.

Couldn't the type checking be done after the function was resolved?

Stumblor:
Ok - I'm coming around more to this way of thinking. Going to close this now.

PR #1055, opened 2017-11-11
Updated to explain getters in more detail

myaroshefsky:
It can be confusing that getters are defined as functions but accessed as properties.  This is explained in greater detail. 

It also can be confusing how getters can be accessed using Vue templates.  Explanation is provided why it is necessary to mapGetters


LinusBorg:
I don't think that should be necessary, at least not in this form. 

As already hinted to in the paragraph before the ones you added, getters are like computed properties - to be precise, the *are* computed properties, so they work like computed properties (which people using vuex should already know about from the core docs).

If people understand that, then you don't have to explain about JavaScript getters etc. 

Do maybe we should stress more that getters are computed properties?

kiaking:
Closing the PR due to inactivity. Please open an issue if we need to discuss about this further 👍

Issue #1037, opened 2017-11-01
mapSetters (similar to mapGetters) for adding setters to mutations


Shashwat986:
### What problem does this feature solve?
There have been lots of situations where I have some variables in `state` that I need to assign values to. The correct way to modify the state tree is to create a mutation to set the value of the key in the state. This means a lot of repetitive code is written (as shown below):

```
export default {
  state: {
    name: null,
    value: 0
  },
  mutations: {
    setName (state, name) {
      state.name = name;
    },
    setValue (state, value) {
      state.value = value;
    }
  }
};

$store.commit('setName', 'John Doe');
```

`mapSetters` will allow users to create all these setters using a single like of code.

### What does the proposed API look like?
```
import { mapSetters } from 'vuex'

export default {
  state: {
    name: null,
    value: 0
  },
  mutations: {
    ...mapSetters(['name', 'value']),
    //...
  }
};
```

Usage: `$store.commit('name', "John Doe")` OR `$store.commit('setValue', 4)` depending on how it is implemented. I prefer the second approach myself.

Shashwat986:
I'm interested in working on this myself, so no effort beyond approval is required from anyone else's end.

thebrubaker:
I wrote a simple function where I "bootstrap" the mutations and getters on my module, which might be useful to you:

``` js
export function bootstrap (module) {
  if (module.getters === undefined) {
    module.getters = {}
  }

  if (module.mutations === undefined) {
    module.mutations = {}
  }

  Object.keys(module.state).forEach(key => {
    if (module.getters[key] === undefined) {
      module.getters[key] = state => state[key]
    }
    if (module.mutations[key] === undefined) {
      module.mutations[key] = (state, value) => {
        state[key] = value
      }
    }
  })

  return module
}
```

``` js
import { bootstrap } from '~/utilities/vuex'

export default bootstrap({
  namespaced: true,
  state: {
    name: '',
    email: ''
  }
})
```

Shashwat986:
Exactly, something like this should be part of the actual vuex package.

thebrubaker:
Yup you could do that just as well. It certainly wouldn't hurt to add some more utilities, but at the end of the day it's also pretty trivial to implement it yourself. Who knows!

darrinmn9:
[...]

PR #1060, opened 2017-11-14
Add mapStoreSetters and mapStoreGetters helper

Shashwat986:
Adding the feature discussed in #1037 

So, I'm not entirely sure about this commit and how it's to be deployed?
`/dist` doesn't seem to be in `.gitignore` so I've added it to the commit, but that seems a little weird. Additionally, I'm not too sure about the changes I've made in `src/index.esm.js` either.

If any further changes are required, please feel free to point them out


ktsn:
See: https://github.com/vuejs/vuex/issues/1037#issuecomment-344448231

PR #1061, opened 2017-11-18
in the example at the very bottom, use dispatch

mikhailvs:
The commit method should be avoided in from inside components.


ktsn:
I think we can use mutation directly in this case because it is used for form binding intentionally. We probably don't do the other than assigning value in the setter.

ktsn:
Closing for the above reason. Thanks for the PR though!

PR #1067, opened 2017-11-25
Add subscribeAction in index.d.ts fixes #1066

### LINKED ISSUE #1066: TypeScript declaration for subscribeAction missing
### Version
3.0.1

### Reproduction link
[https://github.com/vuejs/vuex/blob/dev/types/index.d.ts](https://github.com/vuejs/vuex/blob/dev/types/index.d.ts)

### Steps to reproduce
Try to subscribe to an action within the store with the new introduced method in version 3.0.1 subscribeAction.

### What is expected?
No error

[ts] Property 'subscribeAction' does not exist on type 'Store<{}>'.

---
Please add typescript declaration just like line 19 of https://github.com/vuejs/vuex/blob/dev/types/index.d.ts
subscribeAction<P extends MutationPayload>(fn: (mutation: P, state: S) => any): () => void;

and instead of MutationPayload maybe ActionPayload.
and instead of mutation maybe action

snaptopixel:
Looks like this one slipped through the cracks, thanks for all the hard work!

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -17,6 +17,7 @@ export declare class Store<S> {
   commit: Commit;
 
   subscribe<P extends MutationPayload>(fn: (mutation: P, state: S) => any): () => void;
+  subscribeAction<P extends ActionPayload>(fn: (action: P, state: S) => any): () => void;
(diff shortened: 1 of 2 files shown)


ktsn:
Thanks. Could you add a test for this?

snaptopixel:
Should be all set @ktsn, thanks!

panxiuqing:
Hope release a version with this fix

PR #1074, opened 2017-12-03
Update testing.md

Keaws:
Word 'can' implies the user has a choice to do that or not, but in this case it is obligatory to export the mutations as a named export. Took me a while to make the test work :)


ktsn:
Thanks!

PR #1088, opened 2017-12-13
Add support for mixins

### LINKED ISSUE #1078: Vuex Mixins
Once a Vuex module grow is recommended to divide it into several files composing the different sections, but then it gets interesting to reuse code. For example, there could be several modules dealing with DATA in an exact same way, but not using all the features: one requires selecting data, another requires specific API actions, etc.

A way to solve this is to divide the module into even smaller reusable parts, Data, DataSelection, DataApi, etc. and combine as required when building the module:

```js
import data from './common/data'
import selection from './common/data-selection'

export default {
  state: () => ({
    ...data.state,
    ...selection.state
  }),
  getters: {
    ...data.getters,
    ...selection.getters
  },
  mutations: {
    ...data.mutations,
    ...selection.mutations
  }
}
```

You may suggest setting those mixins as independent modules, but that is not convenient because modules force the state into objects (state.selection.selection vs state.selection) and adds extra layer of complications when trying to access the state from other mixin.

Instead of the above, a mixin would allow doing:

```js
import data from './common/data'
import selection from './common/data-selection'

export default {
  mixins: [data, selection]
}
```

If this idea doesn't align up with the core base, it would be nice to at least expose some way to allow adding this functionality through a plugin.

manico:
Fixes #1078
If this is ok direction, I can create tests.


src/store.js
```diff
@@ -78,6 +80,17 @@ export class Store {
     }
   }
 
+  applyMixins (options) {
+    if (Array.isArray(options.mixins)) {
+      const merge = Vue.config.optionMergeStrategies.computed
```
ktsn: I think it is a bad idea to rely on the merge strategy in Vue core because it can easily be broken if the strategy changes the behavior. Let's write it in the Vuex code.
Also, I suspect if we should use `computed` merge strategy since it overwrites a property if there are the same key in the objects. I feel it would be natural to take lifecycle hook merge strategy.

src/store.js
```diff
@@ -25,6 +25,8 @@ export class Store {
       strict = false
     } = options
 
+    this.applyMixins(options)
```
ktsn: This applies mixins only root options. We should even apply them for all child modules.

src/store.js
```diff
@@ -78,6 +80,17 @@ export class Store {
     }
   }
 
+  applyMixins (options) {
+    if (Array.isArray(options.mixins)) {
+      const merge = Vue.config.optionMergeStrategies.computed
+      options.mixins.forEach(mixin => {
+        for (const key in mixin) {
+          options[key] = merge(mixin[key], options[key])
```
ktsn: Since the `options` object is mutated, there would be a problem if the module is reused in multiple times. i.e. it reapplies the mixins for the already applied options object.

src/store.js
```diff
@@ -78,6 +80,17 @@ export class Store {
     }
   }
 
+  applyMixins (options) {
```
ktsn: It would more appropriate to put this logic in https://github.com/vuejs/vuex/blob/dev/src/module/module.js. Because `Module` class is responsible for each module.

manico:
Sure. Just wanted to first discuss if we should implement this. If so, will do appropriate changes and tests. Thanks for reply.

manico:
Added simple merge strategy to add property from mixin if not present.
I think that would be sufficient for store as opposed to vue core where different strategies are applied.
We could maybe chain actions, but I'm not sure. 🤔

kiaking:
Closing the issue due to its age, and also by the closed #1078.

PR #1090, opened 2017-12-14
Add isDebug util

manico:
On a lot of places in code we have NODE_ENV check to see if it is production env. This PR introduces `isDebug` util function that does the same, but it is a lot cleaner and shorter.


ktsn:
Thanks for the contribution. However, writing the `NODE_ENV` check as inline is intentional. If we extract it with some function, it will no longer eliminate in minification process.

PR #1099, opened 2017-12-21
 Add Simple test case for base function utils

lanzhiheng:


lanzhiheng:
@Jinjiang OK，I will remove it.

test/unit/util.spec.js
```diff
@@ -1,6 +1,11 @@
-import { deepCopy } from '../../src/util'
+import { find, deepCopy, forEachValue, isObject, isPromise, assert } from '../../src/util'
 
 describe('util', () => {
+  it('find', () => {
+    const list = [33, 22, 112, 222, 43]
+    expect(find(list, function (a) { return a % 2 }))
```
ktsn: Sorry for reviewing after approved this but this does not specify an expected value. It looks always pass the test.

lanzhiheng: 😢 Sorry, I have updated it. @ktsn

PR #1103, opened 2017-12-26
Refactoring: Use isFunction and isString utility

manico:
This PR removes repetitive type checks for function and string by introducing simple `isFunction` and `isString` utility functions.

There is a small change in dist size.

Before:
dist/vuex.js 24.94kb
dist/vuex.min.js 9.32kb (gzipped: 2.94kb)
dist/vuex.common.js 25.33kb
dist/vuex.esm.js 25.42kb

After:
dist/vuex.js 24.93kb
dist/vuex.min.js 9.29kb (gzipped: 2.98kb)
dist/vuex.common.js 25.33kb
dist/vuex.esm.js 25.42kb


ktsn:
Hmm, I don't think we need such helper since using `typeof` is short enough. We have `isObject` because it needs `null` check additionally. 
I'm concerning an extra maintenance cost. (i.e. should we add such helpers if we want to check other types?)

manico:
Reasoning is to add helper if we use it on multiple places and it shortens code.
But, I see your point, so feel free to close PR.

ktsn:
Reducing just 0.03kB with including extra maintenance cost would not considerable trade off, IMO.
Thank you for trying to improve Vuex, though 🙂

PR #1104, opened 2017-12-26
[Shopping cart example] Use `product` instead of abbreviation

hootlex:
Replace the `p` variable name abbreviation with `product` to make the example easier to read.


hootlex:
https://github.com/vuejs/vuex/pull/1107

PR #1105, opened 2017-12-26
Use keyof to improve type inference in helpers

Demivan:
Using map helpers (`mapState`, `mapMutations`, `mapGetters` or `mapActions`) breaks type checking in Vue components.
This pull request uses `keyof` Typescript 2.1 feature to fix this.
Before change:
beforechangethis
beforechange
After change:
afterchangethis
afterchange
Breaking change: When using functions as mappings, first parameter (`Commit`, `Dispatch`) type is not inferred, but it is type-checked only.


ktsn:
Thank you for the contribution. However I'm actually just making a major update for helper types which includes the fix to this problem and improvements to use strictly typed getters/actions/mutations. (See #1121)

Sorry for the inconvenience. I hope you did not mind to use #1121 as the fix for this problem.

Demivan:
Glad to hear that. https://github.com/vuejs/vuex/pull/1121 works fine for me.
Closing this pull request

PR #1108, opened 2017-12-29
Avoid updating license every year

m1guelpf:
Uses the same template as in main repo to avoid recieving a PR every year.


ktsn:
Thanks!

PR #1110, opened 2018-01-01
Update Copyright 2018

edwinychu:


ktsn:
Thanks for PR. But we already had the same kind of PR before this one #1108 🙂

PR #1113, opened 2018-01-02
Add some test and comments for module.js

lanzhiheng:


test/unit/module/module.spec.js
```diff
@@ -21,9 +21,156 @@ describe('Module', () => {
...
+    module.addChild('v2', new Module({}))
+    expect(Object.keys(module._children)).toEqual(['v1', 'v2'])
+  })
+
+  it('remove child method', () => {
+    const module = new Module({})
+
+    module.addChild('v1', new Module({}))
+    module.addChild('v2', new Module({}))
+    module.removeChild('v2')
```
ktsn: We should also check the `_children` just before calling `removeChild` to confirm whether it actually works or not.

test/unit/module/module.spec.js
```diff
@@ -21,9 +21,156 @@ describe('Module', () => {
...
+        name: 'vuex',
+        version: '3.x.x'
+      }
+    }
+    const module = new Module(originObject)
+
+    expect(module._rawModule).toEqual(originObject)
+
+    module.update(newObject)
+    expect(module._rawModule).not.toEqual(newObject)
```
ktsn: I think this assertion is not needed because it does not matter whether the whole object strictly equals `newObject` or not.

lanzhiheng: @ktsn Thank you for your comment, I will fix it today.

PR #1121, opened 2018-01-04
feat: improve helper types for more type safety

### LINKED ISSUE #532: Typescript typed vuex
Right now it's hard to maintain type safety with `Vuex` and `Typescript` because of the way the `$store` exposes actions and mutations through string-based `dispatch` and `commit`. At compile time there's no way to know the strings are valid or that the passed payload has the requisite fields.

There's this valiant effort:

- https://github.com/HerringtonDarkholme/kilimanjaro
- https://herringtondarkholme.github.io/2016/10/03/vue2-ts2/

Though that API doesn't really look like the `Vuex` API to me. Would there be interest in trying to allow typescript type safety in Vuex? Are there existing best practices?

### LINKED ISSUE #564: Write Stores in Typescript
I couldn't find anything on google, so is this conveniently possible? Basically the bummer for me is that I have to call `dispatch`/`commit` with a given identifier string to trigger something. 

My vision is to define stores as Typescript classes with typed methods (getting intellisense in editor and so on) which I can call instead of `dispatch("my_action", data)` and having to look up each and everything and check for errors manually all the time. 

Basically my problem is that my team is building a fairly large Vue/Vuex frontend for a core product of our company, consisting of already 18 fairly complex Stores as modules, and this year stuff is going to at least quadruple in size and complexity. We already decided to go on with typescript instead of ES7 and are in the process of migrating our backend/frontend code, but this thing really feels like a bummer. 

I think this is not the typical "small" use case for Vue.js, since we're building a huge enterprise frontend with loads of requirements and edge cases, but isn't vuex(/flux) supposed to scale up when complexity rises? 

Has anyone experience in building complex, type safe Vuex 2.0 stores in Typescript yet? Any help here would be appreciated

### LINKED ISSUE #1119: Type checking is broken in Vue components when using mapping helpers
### Version
3.0.1

### Reproduction link
[https://github.com/vuejs/vuex/pull/1105](https://github.com/vuejs/vuex/pull/1105)

### Steps to reproduce
1. Create new Vue component.
2. In jsconfig.json (tsconfig.json) enable strict mode and js checking
3. Add `mapState` or `mapActions` to the component

### What is expected?
Type checking should work.
Intellisense should suggest actions mapped using `mapActions` and suggest data mapped using `mapState`
image

Type checking stops working.
Intellisense is not suggesting mapped actions and state.
image

---
This is due to `mapState` and `mapActions` returning `Dictionary<any>`
I created pull request that fixes this for me: https://github.com/vuejs/vuex/pull/1105

ktsn:
This typings update allows to use typed getters/actions/mutations out of the box if they are used in the following manner.

**1. Declare each store assets types as interfaces.**

```ts
// State
export interface CounterState {
  count: number
}

// Getters
// key: getter name
// value: return type of getter
export interface CounterGetters {
  power: number
}

// Mutations
// key: mutation name
// value: payload type of mutation
export interface CounterMutations {
  increment: { amount: number }
}

// Actions
// key: action name
// value: payload type of action
export interface CounterActions {
  incrementAsync: { amount: number, delay: number }
}
```

**2. annotate a namespaced module with `DefineModule` utility type.**

The annotated assets must fulfill specified names and return type (getters) / payload type (actions, mutations). Also the assets types will be inferred.

The type in the following example should be fully inferred:

```ts
import { DefineModule } from 'vuex'

const counter: DefineModule<CounterState, CounterGetters, CounterMutations, CounterActions> = {
  namespaced: true,

  state: {
    count: 0
  },

  getters: {
    power: state => state.count * state.count
  },

  mutations: {
    increment (state, payload) {
      state.count += payload.amount
    }
  },

  actions: {
    incrementAsync ({ commit }, payload) {
      setTimeout(() => {
        commit('increment', { amount: payload.amount })
      }, payload.delay)
    }
  }
}
```

**3. create typed namespaced helpers with `createNamespacedHelpers`.**

Then, we can acquire typed mapXXX helpers for the defined namespaced module.

```ts
export const counterHelpers = createNamespacedHelpers<CounterState, CounterGetters, CounterMutations, CounterActions>('counter')
```

**4. use the namespaced helpers in a component.**

```ts
import { counterHelpers } from '@/store/modules/counter'

export default Vue.extend({
  computed: counterHelpers.mapState({
    value: 'count'
  }),

  methods: counterHelpers.mapMutations({
    inc: 'increment'
  }),

  created () {
    // These are correctly typed!
    this.inc({ amount: 1 })
    console.log(this.value)
  }
})
```

### Store is still not typed

I think it is probably impossible to infer the entire store type correctly since we cannot concat getter/actions/mutations names with namespace on type level. So this PR focuses how we do not use `$store` directly but use typed helpers instead.

### It does not infer the types completely if passing functions to mapXXX helpers.

For example:
```ts
counterHelpers.mapState({
  foo: state => state.count
})

counterHelpers.mapMutations({
  bar (commit) {
    commit('increment', { amount: 1 })
  }
})
```

We can write the above code with inferred `state` and `commit` types but the component will have a type of `foo: any` and `bar: (...args: any[]) => any`.

It can be easily rewrite with a combination of object form mapXXX helpers and normal methods, so I think it would not be a problem.

### Using root assets

The default mapXXX helpers still accepts any asset names and returns record of `any` type. To manually annotate them, we can use `createNamespacedHelpers` too. If we don't specify a namespace as an argument, it returns the root mapXXX helpers so that we annotate with root module assets types as same as namespaced modules.

```ts
const rootHelpers = createNamespacedHelpers<RootState, RootGetters, RootMutations, RootActions>()
```

fix #532 
fix #564 
fix #1119


types/index.d.ts
```diff
@@ -36,43 +48,89 @@ export declare class Store<S> {
...
+  rootGetters: RG;
 }
 
-export interface Payload {
+export interface BasePayload {
   type: string;
 }
 
-export interface MutationPayload extends Payload {
+type Payload<K extends keyof P, P> = { type: K } & P[K]
```
HerringtonDarkholme: Technically this is a breaking change. I don't know how frequent `Payload` is used, but this might be a compatible alternative:

```ts
type Payload<P = { [k: string]: {} }, K extends keyof P = keyof P> =
    { type: K } & P[K]
```

http://www.typescriptlang.org/play/#src=type%20Payload%3CP%20%3D%20%7B%20%5Bk%3A%20string%5D%3A%20%7B%7D%20%7D%2C%20K%20extends%20keyof%20P%20%3D%20keyof%20P%3E%20%3D%0A%20%20%20%20%7B%20type%3A%20K%20%7D%20%26%20P%5BK%5D%0A%0Adeclare%20var%20p%3A%20Payload%0A

HerringtonDarkholme:
Thanks! This is a long change so I think I need some time to review the change. Also, we need documentation update to catch up this new awesomeness. 😄

types/helpers.d.ts
```diff
@@ -1,69 +1,100 @@
...
-interface Mapper<R> {
-  (map: string[]): Dictionary<R>;
-  (map: Dictionary<string>): Dictionary<R>;
+/**
+ * Utility types to declare helper types
+ */
+type Computed<R> = () => R;
+type Method<R> = (...args: any[]) => R;
+type MutationMethod<P> = (payload: P) => void;
+type ActionMethod<P> = (payload: P) => Promise<any>;
```
HerringtonDarkholme: I'm fine with untyped action return type. In fact, users can register multiple actions so the return type can't be checked any way.

types/helpers.d.ts
```diff
@@ -1,69 +1,100 @@
...
+  <S extends State = State, G extends Getters = Getters, Map extends BaseStateMap<S, G> = BaseStateMap<S, G>>(namespace: string, map: Map): { [K in keyof Map]: Computed<any> };
 }
 
-interface FunctionMapper<F, R> {
-  (map: Dictionary<(this: CustomVue, fn: F, ...args: any[]) => any>): Dictionary<R>;
+/**
+ * mapMutations
+ */
+interface MapMutations<Mutations> {
+  <M extends Mutations = Mutations, Key extends keyof M = keyof M>(map: Key[]): { [K in Key]: MutationMethod<M[K]> };
```
HerringtonDarkholme: Usually generic default is for interface / type alias. Mapper is not exported nor augmented by users, so I think these defaults aren't required.

types/helpers.d.ts
```diff
@@ -1,69 +1,100 @@
...
+  mapGetters: MapGetters<Getters>;
+  mapMutations: MapMutations<Mutations>;
+  mapActions: MapActions<Actions>;
 }
 
-export declare const mapState: Mapper<Computed>
-  & MapperWithNamespace<Computed>
-  & MapperForState
-  & MapperForStateWithNamespace;
+export declare const mapState: RootMapState<BaseType, BaseType>;
```
HerringtonDarkholme: The proposed usage is `mapState<RootState>(['foo', 'bar'])`, however, the return type is `RootState` because `foo/bar` isn't participating inference.

Another usage is that we export  `RootMapXXX` and slightly change their type.

```ts
export interface RootMapState<State> {
  <Keys extends keyof State>(keys: Keys[]): {[K in Keys]: State[K]}
}
```

then users can write something like `const myMapState: RootMapState<RootState> = mapState`.  Of course, it requires us to maintain more public types, and also looks a little bit bizarre.  But it is more versatile in usage -- I think object style can be supported, and more precise in returning type.

HerringtonDarkholme: How about reusing the pattern of `createNameSpacedHelpers`? Adding one more helper function like `createRootHelpers`? Implementation wise, it just return these root helpers, but it gives better developer experience for both end users and lib maintainers.

types/utils.d.ts
```diff
@@ -0,0 +1,79 @@
...
+
+/**
+ * Infer getters object type from passed generic types.
+ * `Getters` is an object type that the keys indicate getter names and
+ * its corresponding values are return types of the getters.
+ * `State` is a module state type which is accessible in the getters.
+ * `ExtraGetters` is like `Getters` type but will be not defined in the infered getters object.
+ * `RootState` and `RootGetters` are the root module's state and getters type.
+ */
+export type DefineGetters<
```
HerringtonDarkholme: What about a one-for-all option? like : 
```ts

export type DefineStore<
  Actions,
  State,
  Getters,
  Mutations,
  ExtraActions = {},
  RootState = {},
  RootGetters = {},
  RootMutations = {},
  RootActions = {}
> = {
  // ....
}
```

types/helpers.d.ts
ktsn: > How about reusing the pattern of createNameSpacedHelpers

That sounds good idea. I think returning root helpers from `createNamespacedHelpers` with no argument would work. `createRootHelpers` would sound a bit weird for pure JS users.

types/utils.d.ts
ktsn: Yes, it looks better since we always combine all assets for module 😄 
I will rewrite the utility types.

types/helpers.d.ts
HerringtonDarkholme: Awesome! Great balance between types and runtime behavior.

types/helpers.d.ts
ktsn: Actually, it throws error if `Mutations` is `BaseType` since `M` will be inferred as `{}` in that case. I have no idea how we fix it without default generic type 😞 

But we may no longer need it if we create typed root helpers with `createNamespacedHelpers` in any cases.

ktsn:
> documentation update

I will write docs for this. Thanks for pointing it out!

types/index.d.ts
ktsn: I ended up reverting the rename of `Payload` https://github.com/vuejs/vuex/pull/1121/commits/0858c6db067259130d5a74210f10eb5cec7bdb43

types/helpers.d.ts
HerringtonDarkholme: Let's try typed root helpers! It looks more promising since it can return more precise types.

types/index.d.ts
HerringtonDarkholme: Yes, changing `interface` to `type` would be a breaking change. My mistake 😷

types/helpers.d.ts
Demivan: There is a problem with returning `MutationMethod` here.
`MutationMethod` does not have `...args: any[]` parameters anymore.
This makes all mutations take one argument, same goes for actions.

ktsn: It's not actually a problem because mutations and actions only accept one parameter. The returned methods will never process 2nd or latter parameters.

Demivan: image

Demivan: You can pass multiple parameters (or none) to mapped actions

ktsn: Ah, I see. It should be an optional argument in default to follow the existing behavior.

But I think if we annotate types via `createNamespacedHelper` explicitly, it should not be an optional. Because it easily breaks apps if the user adds a payload for actions/mutations with no payload and the type checker cannot detect that.

If the users want to declare non-payload actions/mutations, they should explicitly pass a `null` or `undefined` value.

@HerringtonDarkholme What do you think?

types/helpers.d.ts
```diff
@@ -51,50 +69,50 @@ interface RootMapState<State, Getters> extends MapState<State, Getters> {
 /**
  * mapMutations
  */
-interface MapMutations<Mutations> {
-  <Key extends keyof Mutations>(map: Key[]): { [K in Key]: MutationMethod<Mutations[K]> };
-  <Map extends Record<string, keyof Mutations>>(map: Map): { [K in keyof Map]: MutationMethod<Mutations[Map[K]]> };
+interface MapMutations<Mutations, Type extends MethodType> {
+  <Key extends keyof Mutations>(map: Key[]): { [K in Key]: MutationMethod<Mutations[K], Type> };
```
HerringtonDarkholme: This typing still requires all methods in mutation/action are homogeneous: all methods either require one parameter at the same time or don't accept parameter at all. We cannot declare such mutations that some methods require parameter while others don't at the same time.

types/helpers.d.ts
HerringtonDarkholme: I agree that users should explicitly pass `null` or `undefined`.

docs/en/api.md
```diff
@@ -212,6 +212,8 @@ const store = new Vuex.Store({ ...options })
 
   The first argument can optionally be a namespace string. [Details](modules.md#binding-helpers-with-namespace)
 
-- **`createNamespacedHelpers(namespace: string): Object`**
+- **`createNamespacedHelpers(namespace?: string): Object`**
 
   Create namespaced component binding helpers. The returned object contains `mapState`, `mapGetters`, `mapActions` and `mapMutations` that are bound with the given namespace. [Details](modules.md#binding-helpers-with-namespace)
+
+  If the namespace is not specified, it returns the root mapXXX helpers. This behavior is convenient to annotate strict types for mapXXX helpers.
```
HerringtonDarkholme: Hmmm, what about 

> This is mainly for TypeScript users to annotate root helper's type.

Mentioning TypeScript explicitly so JS users know annotating type doesn't require much care for them.

types/index.d.ts
```diff
@@ -12,9 +12,7 @@ export {
 } from "./helpers";
 
 export {
-  DefineGetters,
```
HerringtonDarkholme: IMHO it doesn't bother if we re-export other utility type helpers. These helper types can be convenient for users to separate getters/actions/mutations to different files.

types/helpers.d.ts
blake-newman: I agree, i believe `undefined` is probably best to signal that this payload is void. `null` could have context to an application where undefined can't be used on the store as it breaks the reactivity principles.

types/helpers.d.ts
ktsn: Yes, I intend that behavior. I leave them optional if the users do not annotate types because they probably want flexible syntax like in JS. On the other hand, the methods always require an argument if they annotate types because they probably want type safety in that case.

ktsn:
@blake-newman I believe this PR does not include any breaking changes on the user land while it is large change of the internal. So we probably can release this improvement as minor update.
Did you find any breakages caused by this change?

docs/en/api.md
ktsn: It sounds clearer than before. Thanks!

types/index.d.ts
ktsn: Makes sense. I've exposed them again.

blake-newman:
@ktsn Nope no breaking, however due to this being a large improvement around TS. To me it makes sense to add this in while we are doing the other changes. As it will require visiting the documentation to understand the new typing available.

We can proceed without this being breaking, it just a matter of justification. Should we take the opportunity to add a breaking change to further aid the improvements? @yyx990803 What is your preference?

ktsn:
I actually don't mind whether this change is released as major or minor bump, will second @yyx990803's decision.

BTW, for anyone using class-style components with TS, I'm also working on it: https://github.com/ktsn/vuex-class/pull/16

ktsn:
Made `dispatch` and `commit` in annotated modules more strict as same as mapped actions/mutations - they now always requires payload so that they ensure the payload cannot be invalid in run time thanks to type checking.

docs/en/typescript.md
```diff
@@ -0,0 +1,228 @@
...
+
+`createNamespacedHelpers` is made for generating new component binding helpers focusing a namespaced module. The API however is useful to create typed root binding helpers. So if you need them, you call `createNamespacedHelpers` without passing namespace:
+
+```ts
+const rootHelpers = createNamespacedHelpers<RootState, RootGetters, RootMutations, RootActions>()
+```
+
+## Explicit Payload
+
+While regular (not strictly typed) `dispatch` and `commit` can omit a payload, typed ones does not allow to omit it. This is because to ensure type safety of a payload. If you want to declare actions / mutations that do not have a payload you should explicitly pass `undefined` value.
```
HerringtonDarkholme: We can add a note telling users this might change in future. https://github.com/Microsoft/TypeScript/pull/21316

ktsn: 👍 
I'll add it later.

bpicolo:
Seems rather boilerplate heavy. Is there an imaginable version of this that doesn't require separately defined interfaces do you think?

ktsn:
I'm not sure if we can eliminate separated interface declarations with well type inference/safety in the current Vuex API. But I think this approach is balanced because we should write such module types only once. That means we do not need to write type annotations in a module implementation and for mapped properties/methods. All things will be inferred from the interface types which originally declared. So I think it's not so much heavy in total.

bpicolo:
@ktsn as far as defining $store, what would it look like to also define an explicit interface type for store? 

Another PoC - looks like https://github.com/mrcrowl/vuex-typex uses a builder pattern to generate a typed store.

tvkit:
I included the _verified_ commit c3626f7 into my project, and it's working well. Thank you!

After some trial-n-error, I found that I could only invoke actions via the `$store.dispatch` method. Should I expect mapped methods on `this`?

I adopted the `$store` structure provided in `vuex/types/test/shopping-cart`, but I'm not sure about action integration.

Regarding the shopping cart example, I'm curious how the `getAllProducts` method would be invoked:

```
export interface ProductsActions {
  getAllProducts: undefined
}

```
and possibly with the addition of something like:

```
export interface ProductsActions {
  getAllProducts: undefined;
  save(t: T, q: Q);
}

```

ktsn:
As I already stated in the description of this PR, this typing does not declare `$store`. Always use mapped methods.

tvkit:
Thanks, @ktsn.  I overlooked step 4 ... and 3 ...

Should it be of any value, two things that helped to get my project going:

- webpack/ts-loader didn't like the spread operators, so I used `Object.assign()` to merge Mutations and Actions. 
- used `MyBase` as a workaround for incorporating `computed:` and `methods:`  -- couldn't get it to work directly within the class.  

```
import { myHelpers } from "@store/modules/my";

const MyBase = Vue.extend({
  computed: myHelpers.mapState(["counter"]),
  methods: Object.assign({},
    myHelpers.mapMutations(["increment"]),
    myHelpers.mapActions(["foo", "bar"]))
});

@Component
export default class MyComponent extends MyBase {

  created() {
    this.foo({ what: "foo" });
    this.bar({});
    this.increment({ amount: 4 });
    this.counter;
  }
}
```

ktsn:
> webpack/ts-loader didn't like the spread operators

I don't understand what it means. spread operators should work you use any bundlers.
Also, `Object.assign` should return merged type of parameters. https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es2015.core.d.ts#L302

> used MyBase as a workaround for incorporating computed: and methods: -- couldn't get it to work directly within the class.

Please read carefully this thread. https://github.com/vuejs/vuex/pull/1121#issuecomment-356983124

tvkit:
The spread issue is related to webpack/ts-loader, and there's the possibily `...` support exists in the recent WebPack 4.0 release, [Spread Operator bug?](https://github.com/webpack/webpack/issues/5548)

Regarding a direct TypeScript compile, tsc (v2.7.2) is not complaining at all. As I described, Object.assign() works perfectly which is great given my existing WebPack config.

> Please read carefully this thread. #1121 (comment)

I previously noticed the npm [vue-class-component](https://www.npmjs.com/package/vue-class-component). Won't integrate that until I reach critical milestones. For now, I'm good with: `class MyComponent extends Vue.extend({awesomeHelpers})`. 

Something worth mentioning for nested modules:

```
// psuedo module layout
const store = {
  module: {
    parent: {
      modules: {
        child
      }
   }
}
```
the child helper could be setup like this:

```
export const childHelpers = 
    createNamespacedHelpers<ChildState, ChildGetters, ChildMutations, ChildActions>("parent/child");
```
where the namespace references the parent.

So up to now, I've builtup a store hierarchy that's working well, and I'm looking forward to the eventual pull so I can avoid this in package.json:
```
  ...
  "dependencies": {
    ...
    "vuex": "https://github.com/vuejs/vuex/archive/c3626f779b8ea902789dd1c4417cb7d7ef09b557.tar.gz",
    ...
  }
```

roblav96:
@ktsn Absolutely wonderful PR! Lobby to merge -> @yyx990803

omidb:
Looks really nice. How does this PR compare to https://github.com/mrcrowl/vuex-typex ?

ktsn:
@omidb `vuex-typex` (and probably other vuex wrapper) provide runtime helper to wrap Vuex and provide other interface to achive type safety. It is the same approach as `vue-class-component`. It may able to achieve complete type safety as the lib auther can modify the API into type checker frendly form but the users need to learn the wrapper API.

On the other hand, this typing update does not affect current runtime semantics in Vuex. That means the users don't need to change their code to make it type safe. They only need to annotate their state/getters/actions/mutations types with type helpers.

blake-newman:
@ktsn since this has been stalled, how much effort do you think this will be to add the conditional types to allow payloads to be omitted?

roblav96:
@ktsn Also haven't found a way to arbitrage `<any>`
https://github.com/vuejs/vuex/blob/f9422a88da3de92d92388b2d0049b5c693c8a91a/types/vue.d.ts#L10
https://github.com/vuejs/vuex/blob/f9422a88da3de92d92388b2d0049b5c693c8a91a/types/vue.d.ts#L16
Current solution:
`"postinstall": "rimraf node_modules/vuex/types/vue.d.ts",` 😆

ktsn:
@blake-newman Actually, I already tried it but it is trickier than I thought at the first glance. I'll look into it deeper when I have sufficient time to do that.

@roblav96 FYI #1192

vahdet:
For the time, is there any way to use that  `import { DefineModule } from 'vuex'` line? Is it available with the [latest dev build](https://vuex.vuejs.org/en/installation.html)?

lzgrzebski:
any ETA when this will be relased? Currently I have to patch my ts code with some explicit 'any' to make it work which isn't ideal.. :<

uoc1691:
Great Work! Question on the usage.
Can the helper be used like  
`storeHelper.mapMutations(["doSomething"]).doSomething({event: new MyCustomEvent()});`

Without doing

`methods: counterHelpers.mapMutations({
    inc: 'increment'
  })`

kirilvit:
Please, resolve the conflicts

uoc1691:
Within an action "this" refers to the store instance. Can we modify DefineActions type like below?

`export type DefineActions<
  Actions,
  State,
  Getters,
  Mutations,
  ExtraActions = {},
  RootState = {},
  RootGetters = {},
  RootMutations = {},
  RootActions = {}
> = {
  [K in keyof Actions]: (this: Store<RootState>,
    ctx: StrictActionContext<State, RootState, Getters, RootGetters, Mutations, RootMutations, Actions & ExtraActions, RootActions>,
    payload: Actions[K]
  ) => Promise<any> | void
}`

ffxsam:
What's holding up this PR from being merged?

Can anyone tell me what a good workaround is, in the mean time? I'm not that fluent in TypeScript.

hobotroid:
I too am would love this PR to get merged.

RehanSaeed:
Looks like @yyx990803 is yet to review this PR.

zhangbobell:
It will make strong sense for this PR been merged.

latel:
strill waiting for this pr

Raiondesu:
Everybody maintaining this repo seems to have forgotten about all type-improving pull requests...

ffxsam:
Could someone _please_ merge this? @yyx990803 @blake-newman @HerringtonDarkholme

kirilvit:
https://medium.com/the-vue-point/plans-for-the-next-iteration-of-vue-js-777ffea6fabf

ffxsam:
@kirilvit That's a great point, I totally didn't think of that.

VSDekar:
But at least a statement would be fair, instead of just saying nothing...

exherb:
any progress?

pikax:
Made a PR https://github.com/ktsn/vuex/pull/1 updating Typescript@3.3.3

joevandyk:
Vue 3 is still a very long ways away, right? If the maintainer doesn't want to merge this, could they close it and explain why? Or suggest what changes need to be made to get this merged in?

pikax:
I think this can't be merge because doesn't work on typescript 3.x (building errors)

ClickerMonkey:
This is a little unrelated, but I've also create new typings for vuex. Similar to this, except instead of 4 interfaces to represent the state/getters/mutations/actions you specify one and it automatically can detect which properties are each:

https://github.com/ClickerMonkey/vuex-typescript-interface

kiaking:
@ktsn Do you think this PR is still maintainable...? Or maybe we should close this one and start over if we really need it 😓

ktsn:
We should go with Vuex 5's TypeScript improvement. Closing.

PR #1128, opened 2018-01-08
Buffer mutations for vue-devtools

bartlomieju:
This PR provides proof-of-concept for buffering mutations for devtools to use. I am not really happy with this code as it relies on setting `store._devtoolsBuffer = undefined` after vue-devtools boots up in order to prevent memory leak. Anyway, thanks to this solution all mutations are inspectable in devtools.
  
Shouldn't be merge until vuejs/vue-devtools#491 is deployed or this code is refactored!

Closes vuejs/vue-devtools#408

I'll be happy to get some feedback

### DIFF
--- a/src/plugins/devtool.js
+++ b/src/plugins/devtool.js
@@ -6,6 +6,9 @@ export default function devtoolPlugin (store) {
   if (!devtoolHook) return
 
   store._devtoolHook = devtoolHook
+  // buffer mutations for devtools
+  // it will later be removed by devtools
+  store._devtoolBuffer = []
 
   devtoolHook.emit('vuex:init', store)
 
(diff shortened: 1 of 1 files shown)


bartlomieju:
I've added additional property `supportsVuexMutation` to global devtools hook to prevent memory leaks in case user uses old version of vue-devtools. It it safe to merge this now without waiting for devtools.

src/plugins/devtool.js
```diff
@@ -15,5 +20,10 @@ export default function devtoolPlugin (store) {
 
   store.subscribe((mutation, state) => {
     devtoolHook.emit('vuex:mutation', mutation, state)
+
+    // if buffer exists push mutation for later use
+    if (Array.isArray(store._devtoolBuffer)) {
```
Akryum: Why not just `if(store._devtoolBuffer) {`?

bartlomieju: Done

michalsnik:
@ktsn can we merge it? :) it would unlock one issue in vue-devtools

PR #1132, opened 2018-01-11
fix(types): allow a function type for root `state` option

ktsn:
This PR allows to pass a function for root `state` option as same as modules in TypeScript.


nulpatrol:
Why not merged?

PR #1133, opened 2018-01-14
fixed actions.md

Neveryu:
这个参数没有传


ktsn:
Thanks for the PR. However, the original example is correct - `mapActions` helper accept only action names.

PR #1134, opened 2018-01-16
Add test for store.js base function

lanzhiheng:


ktsn:
Thanks for the PR. But these functions are tested via other unit tests and I think testing such private function is a bit too much and make the test case hard to maintain.

Issue #1109, opened 2017-12-29
Update Shopping Cart Example


hootlex:
# What problem does this feature solve?
The goal of updating the example is to make it more friendly to newcomers and also demonstrate some best practices.

# What does the proposed API look like?
After reviewing the shopping cart example and discussing it with @chrisvfritz, I’ve started working on possible improvements.

## Summary
- [x] Use separate mutations to update the checkout status and the cart.
- [x] Move/create cart related getters to the `cart` module.
- [x] Remove abbreviations
- [x] Make the example comply with the style guide.
- [x] Make mutations responsible for updating one piece of state to make debugging easier.
- [x] Use an action (`addProductToCart`) to commit separate mutations on `cart` and `products` modules instead of using a shared mutation. 
- [x] Remove mutation types.
- [x] Remove getters that return a piece of the state as is.
- [x] Make modules namespaced
- [x] Rename state properties

### Mutation types
What do you think of removing the mutation types in order to make the example simpler? Mutation types might be beneficial in some cases but by not using them in the example might make it more friendly to newcomers.

### Remove getters that return a piece of state as is
I feel that this pattern might give the wrong impression to newcomers that they always have to use a getter in order to access the state.
Example: 
```js
const getters = {
  allProducts: state => state.all
}
```

### Other
I think that we could use more descriptive names in Vuex assets. 
For instance `added` can be `cartItems` or just `items` since it’s being accessed as `cart.items`. In general, I like the `items` convention for all modules. This way I always know the property name and distinguish the content based on the module name. For example `users.items`, `products.items​`. 

Also, `addToCart` could be `addProductToCart` since a cart can also contain other things.

That might be too peaky but I believe that when people go over an official example it’s very likely to use it as a reference for best practices.

What do you think? 🙂

chrisvfritz:
> Implemented changes #1107

Just reviewed (and approved) that PR! Great work!

### Proposed changes

> Make the example comply with the style guide.

👍 

> Make mutations responsible for updating one piece of state to make debugging easier.

Yes please! 😂 

> Use an action (`addProductToCart`) to commit separate mutations on cart and products modules instead of using a shared mutation.

Yes! While shared action/mutation names are kind of a cool feature, I think they can make debugging quite difficult.

> Remove getters that return a piece of state as is

Sounds great. Your reasoning makes sense.

[...]

PR #1137, opened 2018-01-18
Shopping cart example update

hootlex:
- remove mutation type
- rename `addToCart` action to `addProductToCart`
- rename products module’s `addToCart` mutation to `decrementProductInventory`
- rename cart module’s `addToCart` mutation to `pushProductToCart`
- make mutations responsible for updating one piece of the state

Now instead of having two mutations of the same type, the `addProductToCart` action commits `pushProductToCart` and `decrementProductInventory`.

cc @ktsn https://github.com/vuejs/vuex/issues/1109


examples/shopping-cart/store/modules/cart.js
```diff
@@ -34,41 +33,53 @@ const getters = {
...
-    }
+  pushProductToCart (state, { id }) {
+    state.added.push({
+      id,
+      quantity: 1
+    })
+  },
+
+  incrementItemQuantity (state, cartItem) {
+    cartItem.quantity++
```
hootlex: @ktsn what do you think of updating the `cartItem` this way, instead of passing the `id` and looking for it in the state?

ktsn: @hootlex I prefer looking for the target item in the mutation as the example. Because the users can pass a non-state object unexpectedly in this way.

hootlex: That's a good point!

hootlex: @ktsn updated accordingly and used a variable to find the item before updating to demonstrate that normal vars can be used to alter the state.

tmaier:
Disclaimer: I am still a beginner when it comes to vuex.

You removed mutation-types. I wonder why. Does this mean the pattern described at [Using Constants for Mutation Types](https://vuex.vuejs.org/en/mutations.html) is not applicable anymore?
I believe an example should always follow the best-practices defined in order to allow easier adoption by beginners.

ktsn:
IMO, mutation constants less make sense now because there are [`namespaced` module](https://vuex.vuejs.org/en/modules.html#namespacing) and upcoming typing improvement (#1121). So it would just enforce the users redundancy rather than its benefit.

Some users who does not use statically typed language may use the constants pattern to avoid typo or for some linting support but I think it should be applied to action types rather than mutation types. Because actions are usually used from entire apps while mutations are from the same module. We may need to update the docs to move the constants section into actions chapter.

vankhuong:
Yeah, actually mutation types are redundant when applied for each module, so docs should be updated TypeScript support soon on the next release

xyzdata:
# where is the `actions.js` & `mutations.js `?

https://vuex.vuejs.org/guide/structure.html

https://github.com/vuejs/vuex/tree/dev/examples/shopping-cart

image

PR #1147, opened 2018-02-01
Proposal: Allow to pass a function as the namespace (to helpers)

mleg:
_Disclaimer_: It's only for duscussion and is not supposed to be merged as it lacks appropriate JSDoc comments, TS-types, and test changes.  

In my app I desperately need to pass a **path** to data from parent to child component **as a prop**.  
But value of the prop is not yet available in the child options when connecting the component to a store.  

**Proposal**: _Allow to pass namespace to helper methods as a function which takes a component instance as its only argument._

When we have in our store a [reusable module](https://vuex.vuejs.org/en/modules.html#module-reuse)  and that module is mostly used by corresponding reusable component such approach seems quite natural to me.  
In parent:
```html
<template>
  <Child storePath="path/toModule" />
</template>
```
With proposed changes we can handle it in child like this:
```javascript
import  { mapState } from 'vuex'
import { property } from 'lodash'

export default {
  props: {
    storePath: String
  },
  computed: {
    ...mapState(property('storePath'), ['someField'])
  }
}
```

Maybe am I missing something and there is some existing simple and obvious solution to the use case?


kiaking:
Closing the PR due to inactivity. The commit log is kinda mess too... Please open another PR if this is something we still need 🤝

PR #1149, opened 2018-02-05
docs(examples): chat example update

keng0o:
- remove mutation type
- use v-model instead of target.value
- use mapAction when using store's action
- use Arrow Function & Object destructuring
- use store/getters

### DIFF
--- a/examples/chat/api/index.js
+++ b/examples/chat/api/index.js
@@ -1,13 +1,13 @@
 const data = require('./mock-data')
 const LATENCY = 16
 
-export const getAllMessages = (cb) => {
+export function getAllMessages (cb) {
   setTimeout(() => {
     cb(data)
   }, LATENCY)
 }
 
-export const createMessage = ({ text, thread }, cb) => {
+export function createMessage ({ text, thread }, cb) {
   const timestamp = Date.now()
   const id = 'm_' + timestamp
   const message = {
--- a/examples/chat/api/index.js
+++ b/examples/chat/api/index.js
@@ -1,13 +1,13 @@
 const data = require('./mock-data')
 const LATENCY = 16
 
-export function getAllMessages (cb) {
+export const getAllMessages = (cb) => {
   setTimeout(() => {
     cb(data)
   }, LATENCY)
 }
 
-export function createMessage ({ text, thread }, cb) {
+export const createMessage = ({ text, thread }, cb) => {
   const timestamp = Date.now()
   const id = 'm_' + timestamp
   const message = {
--- a/examples/chat/components/MessageSection.vue
+++ b/examples/chat/components/MessageSection.vue
@@ -2,13 +2,13 @@
   <div class="message-section">
     <h3 class="message-thread-heading">{{ thread.name }}</h3>
     <ul class="message-list" ref="list">
-      <message 
+      <message
         v-for="message in messages"
         :key="message.id"
         :message="message">
       </message>
     </ul>
-    <textarea 
+    <textarea
       class="message-composer"
       v-model="text"
       @keyup.enter="sendMessage"></textarea>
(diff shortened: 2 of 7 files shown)


examples/chat/components/MessageSection.vue
```diff
@@ -1,54 +1,54 @@
 <template>
   <div class="message-section">
     <h3 class="message-thread-heading">{{ thread.name }}</h3>
-    <ul class="message-list" ref="list">
-      <message
-        v-for="message in sortedMessages"
+    <ul class="message-list"
+      ref="list">
+      <message v-for="message in messages"
```
ktsn: Can we revert the format change? This looks unnecessary and the previous format would readable and generic, IMO.

examples/chat/components/MessageSection.vue
```diff
@@ -1,54 +1,54 @@
...
         :key="message.id"
         :message="message">
       </message>
     </ul>
-    <textarea class="message-composer" @keyup.enter="sendMessage"></textarea>
+    <textarea class="message-composer"
```
ktsn: Add a line break before `class`, please :)

examples/chat/api/index.js
```diff
@@ -1,13 +1,13 @@
...
 
-export function getAllMessages (cb) {
+export const getAllMessages = (cb) => {
   setTimeout(() => {
     cb(data)
   }, LATENCY)
 }
 
-export function createMessage ({ text, thread }, cb) {
+export const createMessage = ({ text, thread }, cb) => {
```
ktsn: Hmm, I don't think we should always use arrow function since function declaration and arrow function is different.

examples/chat/components/ThreadSection.vue
```diff
@@ -6,8 +6,7 @@
       </span>
     </div>
     <ul class="thread-list">
-      <thread
-        v-for="thread in threads"
+      <thread v-for="thread in threads"
```
ktsn: Can we revert this formatting change?

examples/chat/components/ThreadSection.vue
```diff
@@ -19,29 +18,16 @@
...
-    ]),
-    unreadCount () {
-      const threads = this.threads
-      return Object.keys(threads).reduce((count, id) => {
-        return threads[id].lastMessage.isRead
-          ? count
-          : count + 1
-      }, 0)
+  computed: mapGetters(['threads', 'currentThread', 'unreadCount']),
+  methods: mapActions({
```
ktsn: It can be simplified to `mapActions(['switchThread'])`

examples/chat/store/actions.js
```diff
@@ -1,22 +1,17 @@
 import * as api from '../api'
-import * as types from './mutation-types'
 
 export const getAllMessages = ({ commit }) => {
   api.getAllMessages(messages => {
-    commit(types.RECEIVE_ALL, {
-      messages
-    })
+    commit('RECEIVE_ALL', messages)
```
ktsn: Can we change all mutation names in camelCase as same as #1137?

examples/chat/store/getters.js
```diff
@@ -12,3 +12,15 @@ export const currentMessages = state => {
     ? thread.messages.map(id => state.messages[id])
     : []
 }
+
+export const unreadCount = state => {
+  const { threads } = state
```
ktsn: We can move this destructuring assignment in the parameter:

```js
export const unreadCount = ({ threads }) => {
```

examples/chat/store/getters.js
```diff
@@ -12,3 +12,15 @@ export const currentMessages = state => {
...
+  return Object.keys(threads).reduce((count, id) => {
+    return threads[id].lastMessage.isRead ? count : count + 1
+  }, 0)
+}
+
+export const sortedMessages = state => {
+  const messages = currentMessages(state)
```
ktsn: I think using resolved getter is better rather than directly use getter function.

```js
export const sortedMessages = (state, getters) => {
  const messages = getters.currentMessages
```

keng0o:
Thank you for the detailed review
I fixed it, can you check it again?
Thank you.

PR #1164, opened 2018-02-26
it means this?

yanlee26:


LinusBorg:
No, the documentation is correct - the getter receives the state as the first argument.

You only use `this.$store` in components.

PR #1170, opened 2018-03-04
docs: mention mapState needs injected store

blackmiaool:


ktsn:
Hm, I think this description should be stated in Getting Started section since this is not only needed for `mapState` but also for other helper functions and `this.$store` property.

kiaking:
Closing the PR since the store injection explanation is already moved at "Getting Started" section.

PR #1205, opened 2018-03-29
perf: update util find

bichikim:
The find in util doesn't need to loop all value in list for finding only one value


ktsn:
Just my personal thoughts, I prefer the previous version because it's more declarative and shorter.

bichikim:
@ktsn yes you are right and l love code clean and simple however looping all value in list cause a bit performance problem in my opinion.
please refer to this [link](https://github.com/lodash/lodash/blob/master/.internal/baseFindIndex.js)

ktsn:
I don't think this affects actual performance so much. But I don't have strong opinion for this function. I guess I'll merge this change.

PR #1206, opened 2018-03-29
[featrue] dispatch and commit function support multiple parameters

lzxb:
In some cases, multiple parameters need to be passed
### Example
```javascript
store.dispatch('add', 'one', 'two', 'three')
store.commit('add', 'one', 'two', 'three')
```

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -40,10 +40,10 @@ export class Store {
     const store = this
     const { dispatch, commit } = this
     this.dispatch = function boundDispatch (type, payload) {
-      return dispatch.call(store, type, payload)
+      return dispatch.apply(store, arguments)
     }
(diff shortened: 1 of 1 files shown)


bichikim:
I don't like using arguments
way don't you use ...args
`function a(...args){
  using(...args)
}`
it's lovely spread operator (전개 연산자) :)

lzxb:
@bichikim Because it will compile more code

ktsn:
Thanks for contribution. However variadic parameters is intentionally removed since Vuex 2.0. Because it restricts `dispatch`/`commit` usage (e.g. we cannot extend them by passing options object) and not friendly with static typings. You can do the same thing by passing array to the 1st argument.

```js
store.dispatch('add', ['one', 'two', 'three'])
```

PR #1207, opened 2018-03-29
for reading process.env in safe way

bichikim:
I cannot pass "Store should not call root state function twice"
please help me 
process have to be  protected by this way
it is also for improving running speed slightly


ktsn:
Thanks for contribution. However, it is intentional to write `process.env.NODE_ENV` inline since it will be evaluated on compile time so that we eliminate the code for development env by minifier.

PR #1209, opened 2018-03-29
docs: add spies as an alternative to testAction helper

gitlab-winnie:


gitlab-winnie:
@ktsn Could you please have a look if this change makes sense to you?

docs/en/testing.md
```diff
@@ -128,6 +128,24 @@ describe('actions', () => {
...
+
+``` js
+describe('actions', () => {
+  it('getAllProducts', () => {
+    const commit = sinon.spy()
+    const state = {}
+    
+    actions.getAllProducts({ commit, state })
+    
+    expect(dispatch.args).to.deep.equal([
```
ktsn: `commit.args`?

gitlab-winnie: yes, thank you, fixed now 😃

PR #1211, opened 2018-04-02
Update exporting ActionHandler

bichikim:
I should use this type in my project 
And I guess some other people need this as well :)


ktsn:
Thanks. Let's move on #1201 since it supersedes this change 🙂

PR #1214, opened 2018-04-03
[feat] Add asyncData Method Type to ComponentOptions Module Augmentation

andRyanMiller:
We use Vue SSR heavily at @algorithmiaio, and we've followed the [Vue SSR-sanctioned means of declaring async data methods](https://ssr.vuejs.org/en/data.html#logic-collocation-with-components).  We've shimmed together our own module augmentation for the `asyncData` method, but given the prevalence of `asyncData` usage throughout the Vue community, it would be nice to standardize the interface.

I was torn as per where this should live - there isn't a repo for SSR, given that it lives in the Vue repo, and ultimately it felt right here with the Vuex project, since `asyncData` and Vuex go hand in hand. 

I've added a test for the new type as well, and let me know if there's anything I can tweak. Cheers!

### DIFF
--- a/types/test/vue.ts
+++ b/types/test/vue.ts
@@ -8,7 +8,8 @@ const store = new Vuex.Store({
 });
 
 const vm = new Vue({
-  store
+  store,
+  asyncData: ({ store, route }) => new Promise((resolve, reject) => resolve())
 });
 
 vm.$store.state.value;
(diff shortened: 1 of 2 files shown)


ktsn:
Thanks for contribution. However this change is not appropriate for Vuex since `asyncData` is not a part of Vuex API. Also the `asyncData` implementation can be different among implementations and they can conflict if we have this type in Vuex. For example, [Nuxt.js `asyncData`](https://nuxtjs.org/api/context) has its own context type.

Maybe, It should be stated how to augment `asyncData` type in the SSR docs.

andRyanMiller:
Good call - thanks for taking a look!

ktsn:
I just opened an issue on SSR docs vuejs/vue-ssr-docs#174

Issue #1179, opened 2018-03-08
Found text cut off in API Documentation


Toltar:
Wanted to report that in the [api documentation](https://vuex.vuejs.org/en/api.html), there is cut off where it shows the arguments of the capture and dispatch methods. See image below.

![documentationcutsoff](https://user-images.githubusercontent.com/6003536/37168025-310ca088-22c9-11e8-9347-4956d4916da5.PNG)

fnlctrl:
Reopening as this is indeed an issue.

Toltar:
Thanks @fnlctrl , I was unsure of how to report this type of issue on here.

Jinjiang:
Guys how about this solution #1189 ? I have split the long api code into 2 pieces and placed them one-by-one together for workaround.
Thanks.

Jinjiang:
![2018-03-14 9 20 14](https://user-images.githubusercontent.com/206848/37405007-74ec14e6-27ce-11e8-83c7-76a79784eec3.png)

Toltar:
@Jinjiang For a permanent solution, could we provide subsections for each method in the documentation of **Usage** and **Description** for this? Such as:

### Vuex Store Instance Methods
#### `commit()`
- Usage
    - **`commit(type: string, payload?: any, options?: Object)`**
    - **`commit(mutation: Object, options?: Object)`**
- Description
    - Commit a mutation. `options` can have `root: true` that allows to commit root mutations in [namespaced modules.](https://vuex.vuejs.org/en/modules.html#namespacing) [Details](https://vuex.vuejs.org/en/mutations.html)

Jinjiang:
@Toltar Yes it's good.
There already are many docs such as vuex vue-router vue-loader vue-test-utils etc with the same format. It will cost a lot if we decide to make the change. Until now there are just two cases not OK and could be fixed another way easily. Maybe it's not necessary enough at this time imo. 🤔
@posva @chrisvfritz how do you think about it? 🙂
Thanks.

Toltar:
@Jinjiang If it is a lot of work, I am willing to help on changing the overall formatting in the English markdown files. 😄 That is, if everyone agrees to it.

chrisvfritz:
I think keeping one bullet-point per item would be best, but I don't have a problem with splitting it into multiple lines. 🙂

ktsn:
This seems to be resolved by migrating vuepress.

PR #1215, opened 2018-04-05
update for map helper Updated Api

bichikim:
https://github.com/vuejs/vuex/blob/06eb3df66f8690cf59ed57f6cbd16e88695d6837/src/helpers.js#L50
https://github.com/vuejs/vuex/blob/06eb3df66f8690cf59ed57f6cbd16e88695d6837/src/helpers.js#L21

### DIFF
--- a/docs/en/api.md
+++ b/docs/en/api.md
@@ -196,25 +196,25 @@ const store = new Vuex.Store({ ...options })
 
 ### Component Binding Helpers
 
-- **`mapState(namespace?: string, map: Array<string> | Object): Object`**
+- **`mapState(namespace?: string, map: Array<string> | Object<string, string | function(state: any)>): Object`**
 
   Create component computed options that return the sub tree of the Vuex store. [Details](state.md#the-mapstate-helper)
 
   The first argument can optionally be a namespace string. [Details](modules.md#binding-helpers-with-namespace)
 
-- **`mapGetters(namespace?: string, map: Array<string> | Object): Object`**
+- **`mapGetters(namespace?: string, map: Array<string> | Object<string, string>): Object`**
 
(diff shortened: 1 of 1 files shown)


docs/en/api.md
```diff
@@ -196,25 +196,25 @@ const store = new Vuex.Store({ ...options })
...
 
-- **`mapState(namespace?: string, map: Array<string> | Object): Object`**
+- **`mapState(namespace?: string, map: Array<string> | Object<string, string | function(state: any)>): Object`**
 
   Create component computed options that return the sub tree of the Vuex store. [Details](state.md#the-mapstate-helper)
 
   The first argument can optionally be a namespace string. [Details](modules.md#binding-helpers-with-namespace)
 
-- **`mapGetters(namespace?: string, map: Array<string> | Object): Object`**
+- **`mapGetters(namespace?: string, map: Array<string> | Object<string, string>): Object`**
```
ktsn: I guess we can just write it as `Object<string>` since this is a pseudo code and readers probably think the object key is always `string` type?

docs/en/api.md
```diff
@@ -196,25 +196,25 @@ const store = new Vuex.Store({ ...options })
...
 
   Create component computed options that return the evaluated value of a getter. [Details](getters.md#the-mapgetters-helper)
 
   The first argument can optionally be a namespace string. [Details](modules.md#binding-helpers-with-namespace)
 
-- **`mapActions(namespace?: string, map: Array<string> | Object): Object`**
+- **`mapActions(namespace?: string, map: Array<string> | Object<string, string | function(dispatch: function, ...args: any[])>): Object`**
```
ktsn: hmm, it may cause #1179 again. I think we don't need write function argument type here because it is too narrow space to write them. What about just writing `Function` and leave the detail description in its guide section?

bichikim:
You're welcome~!

PR #1216, opened 2018-04-06
[docs] fix grammar

genie-youn:
컴포넌트는 -> 컴포넌트를


ChangJoo-Park:
@ktsn it is typo. @genie-youn is right 👍 

감사합니다

PR #1225, opened 2018-04-10
Update shopping cart example

hootlex:
- Remove getters that return a piece of the state as is
- Make modules namespaced
- Rename cart's `added` to `items`

https://github.com/vuejs/vuex/issues/1109


hootlex:
@ktsn could you review this and the other PRs about updating the examples anytime soon? It'd be nice to have updated examples on Vuex since a lot of people are looking at them for inspiration and best practices 🙂

PR #1226, opened 2018-04-10
Update Counter Hot Example

hootlex:
- Make mutations responsible for updating one piece

The component in this example is named `Counter` (single word). Though, we might want to keep it like that since this is a very simple example. What do you think @ktsn @chrisvfritz ?


examples/counter-hot/store/mutations.js
```diff
@@ -1,9 +1,11 @@
 export const increment = state => {
   state.count++
-  state.history.push('increment')
```
ktsn: IMO, we should leave `state.history.push` along with increment / decrement since they looks atomic. When we increment / decrement `count`, we would always want to update the history.

ktsn:
I think it's ok to name the component `Counter` in this case.

examples/counter-hot/store/mutations.js
chrisvfritz: I'm actually also in favor of keeping this in the mutation, for the same reasons @ktsn offered. By trying to decouple these, we make it possible to create an invalid application state with time travel debugging.

chrisvfritz:
I think if we can maintain simplicity while conforming to the style guide, we should try. What would you both think about `CounterControls`?

examples/counter-hot/store/mutations.js
hootlex: Great catch guys. The idea was to make the actions atomic but atomic mutations in this case is a better approach, especially for time travel debugging.

examples/counter-hot/store/actions.js
```diff
@@ -1,5 +1,10 @@
-export const increment = ({ commit }) => commit('increment')
-export const decrement = ({ commit }) => commit('decrement')
+export const increment = ({ commit }) => {
+  commit('increment')
+  commit('pushToHistory', 'increment')
```
ktsn: Did you forget to remove this line? 😄

hootlex: Whoops! The tests pass locally so I didn't notice 😆

PR #1233, opened 2018-04-15
Problem in safari double id's

wiiikiii:
It seams to be a problem with the app when having id app more than once.


wiiikiii:
it seems to be a local problem. sorry

PR #1240, opened 2018-04-19
update add isProduction function

bichikim:
- change using isProduction instead of process.env.NODE_ENV
vuex frequently uses process.env.NODE_ENV so It needs a isProdunction function


ktsn:
Thanks. But we intentionally write them inline so that bundler pre-evaluate their values in compile time and minifier can eliminate if-blocks.

bichikim:
got it~!

PR #1253, opened 2018-04-28
chore(package.json): add sideEffects: false field to package.json

TheLarkInn:
This PR adds the `"sideEffects": false` property in vuex's `package.json` file. This allow's webpack (for those who want to opt-in to requiring vuex's original source files (instead of the flattened esm bundles), then this will allow webpack to aggressively ignore and treeshake unused exports throughout the module system. 

In many cases this can yield hidden and surprising build time improvements and size reductions in the case that modules flattened from rollup aren't actually needed when webpack performs a scope analysis. 

The angular team has adopted this approach for the CLI yielding some strong build time speed for webpack (since unused exports are also not resolved, parsed, etc.). From my assessment of the module graph created, it looks that there are no sideEffects created against existing re-exports throughout the codebase, therefore even if a user did opt in on their own will, and remove flow types, there should be no danger of removing code that is needed and causing breakages. 

Overall, this already will not be a breaking change in any way but just provides extra benefits to those (and for the cli (if realized) when using webpack 4 and `mode: production`). Users would still have to explicitly alias vuex (as it shows in your examples folder to realize these build speed and size wins). 

Happy to answer any other questions.
~Sean + [webpack team](https://github.com/webpack) 

Related: https://github.com/vuejs/vue/pull/8099

### DIFF
--- a/dist/vuex.common.js
+++ b/dist/vuex.common.js
@@ -1,6 +1,6 @@
 /**
  * vuex v3.0.1
- * (c) 2017 Evan You
+ * (c) 2018 Evan You
  * @license MIT
  */
 'use strict';
@@ -69,6 +69,8 @@ function devtoolPlugin (store) {
  * @param {Function} f
  * @return {*}
  */
+
+
 /**
  * Deep copy the given object considering circular structure.
  * This function caches all nested objects and its copies.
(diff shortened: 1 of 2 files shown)


kiaking:
@TheLarkInn Can you remove the `dist` folder from the commit? Or I'll take over this PR. `"sideEffects": false` is already applied on Vuex 4, so I think it can be merged into Vuex 3 as well.

PR #1254, opened 2018-04-28
Fix: amend the code in eg.

bibi7:
Signed-off-by: wangzhaojie <wangzhaojie@antiy.cn>
hey there,  today i'm learing vuex and i find that a little mistake which may be wrong in this md.
i think use `this.state` instead of `state` may be good cuz the console will not go red
have fun : )


bibi7:
oh..  unware about the argument, my fault, thanks for correct_(:з」∠)_

PR #1271, opened 2018-05-18
Update mutations.md

thoughtkun:
there is no payload **amount**


ktsn:
The current example is correct since it's not actual function call.

PR #1287, opened 2018-05-27
Fix broken link

flaviocopes:
The link to http://vuex.vuejs.org/en/intro.html is a 404, the correct one is https://vuex.vuejs.org/


flaviocopes:
BTW I think it's a good idea to redirect http://vuex.vuejs.org/en/intro.html to https://vuex.vuejs.org/ since I saw that link in multiple places already, including Google searches

PR #1290, opened 2018-05-28
add v-bind:key in v-for

Pewww:
Adding v-bind:key in v-for rendering is recommended.
Also, It makes an error when we run dev server.
So, I added it.
(And, I changed var to let keyword.)


ktsn:
Thanks. But the example is outdated. Please see the one on `dev` branch.

Issue #1162, opened 2018-02-21
Hard-to-locate error when providing wrong parameter type to mapState


joachimboggild:
### Version
3.0.1

### Reproduction link
[https://codesandbox.io/s/048lw4q59n](https://codesandbox.io/s/048lw4q59n)

### Steps to reproduce
1) Open code sandbox provided.
2) Look inside console to the error message provided.

To fix the problem, include brackets inside the mapState statement on line 97 in the HelloWorld.vue file.

### What is expected?
It was expected that an error message would state that the parameter for the mapState function was invalid in 'HelloWorld.vue'.

### What is actually happening?
This error message is returned: 

```
[Vue warn]: Failed to resolve async component: function HelloWorld() {
  return _promise2.default.resolve().then(function () {
    return require("./components/HelloWorld");
  });
}
Reason: TypeError: Cannot convert undefined or null to object
```

---
I just thought that this was a difficult error to track, because the error message provided did not point towards the actual error. If it is possible to make the error message more precise, it would be great.

yawboakye:
@ktsn Current implementation is confident that either an array or object is always passed, which explains the cryptic error & message: it's not stop earlier. What would you expect to be the behavior when you pass something other than that?

EdyHartono:
#1297

sustained:
Hope that PR makes it in, this has bitten me twice so far.

Maybe I'll learn my lesson (if not after this time then) eventually but it will probably keep biting people.

ktsn:
fixed via #1297

Issue #1093, opened 2017-12-15
Should vuex mapState print error message


balazser:
### What problem does this feature solve?
I got the following error: Uncaught TypeError: Cannot convert undefined or null to object
then realized I invoked the  `...mapState(NSP_CONCEPT) `without a second parameter. 
It took some time while I figured out, would have been great, if there is a condition to don't fail or print out a message.

### What does the proposed API look like?
print a message, or put a condition into the function

EdyHartono:
#1297

ktsn:
fixed via #1297

PR #1297, opened 2018-06-03
Should vuex mapState print error message #1093 #1162

EdyHartono:
- return an empty array when map is not valid
- log an error message when map is not valid

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -1,3 +1,5 @@
+import { isObject } from './util'
+
 /**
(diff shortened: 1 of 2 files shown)


jamescostian:
This PR would have saved me an hour today if it were merged. Is there any reason why it hasn't been merged? It's a fairly small and simple PR and it provides Vuex users with much better feedback than is currently provided when you make a mistake with `mapState`

sustained:
Also wondering why this wasn't merged.

PR #1301, opened 2018-06-09
Add parameter types for store's commit and dispatch methods

Curtis017:
This will enable users to provide type safe parameters for the commit and dispatch methods.

Example: 
```javascript
import * as mutations from '@/store/mutation-types';
import uuid from 'uuid'

interface AddTodoRequestPayload {
  title: string;
  text: string;
}

interface AddTodoPayload {
  uid: string
  title: string;
  text: string;
}

const ADD_TODO_REQUEST = 'ADD_TODO_REQUEST';
type ADD_TODO_REQUEST = typeof ADD_TODO_REQUEST;

const actions = {
  [ADD_TODO_REQUEST]({ commit }: { commit: Commit }, payload: AddTodoRequestPayload) {
    const uid: string = uuid();
    commit<mutations.types, AddTodoPayload>(
      mutations.ADD_TODO,
      { uid: uid,  title: 'todo title', text: 'todo text' },
    );
  },
};
```

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -42,6 +42,7 @@ export interface Dispatch {
 }
 
 export interface Commit {
+  <T, P>(type: T, payload?: P, options?: CommitOptions): void;
   (type: string, payload?: any, options?: CommitOptions): void;
   <P extends Payload>(payloadWithType: P, options?: CommitOptions): void;
 }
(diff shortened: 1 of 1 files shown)


kiaking:
I'm not sure if this is really a type safe. I mean yes, it gets typing, but it's providing the types as generics together with the rest of argument. In this case, it wouldn't help much of anything IMO 🤔

kiaking:
Closing due to inactivity. I think we had similar issue but that's closed as well. If there's anything, please open another issue 🤝

PR #1302, opened 2018-06-10
Portuguese Translation

periscuelo:
Hello guys. This time I did a translation with affection looking part by part. Which does not mean that it has no error. So if you can look and let me know if you have something missing, i i'll be glad.


periscuelo:
Hi guys. Have any chance of merge in this month? Thanks! :)

PR #1307, opened 2018-06-14
Fixed API document (Korean) typo

luiseok:
In Korean document(API - Vuex.Store Contructor options part), there was an typo (로투).
It should be fixed as 루트(root).
thanks.


ChangJoo-Park:
@kazupon please merge this PR. :)

Issue #1085, opened 2017-12-08
Easy two-way binding between v-model and vuex #1084


pschaub:
### What problem does this feature solve?
Easy two-way binding between `v-model` and vuex state / mutations.

### What does the proposed API look like?
For example I want to bind the state of the global menu to a vuex state.

Meanwhile I use this manual workaround in my component:
```
computed: {
  nav: {
    get () { return this.$store.state.nav },
    set (v) { this.$store.commit('SET_NAV', v) }
  }
}
```

**It would be helpful to have an easy function like `mapGetters` to do this.**

manico:
I like the idea, just not sure if this is fit for inclusion in library or separate lib.
Here is an example, if it makes sense I can create PR and tests:
https://github.com/manico/vuex/commit/1663b0e11c443e1a22ca3200b9ae18332db62caa

johnmerced:
Looking forward for this! Has this been discussed to see if it should be part of the library or not?

zurcacielos:
Here's a potential solution
https://github.com/maoberlehner/vuex-map-fields

johnmerced:
Any feedbacks yet?

leodutra:
```js
import { mapState } from 'vuex'

// based on article: 
// VueJS - Two way data binding and state management with Vuex and strict mode
// https://ypereirareis.github.io/blog/2017/04/25/vuejs-two-way-data-binding-state-management-vuex-strict-mode/

export default function mapStatesTwoWay (namespace, states, updateCb) {
    const mappedStates = mapState(namespace, states)
    Object.keys(mappedStates)
        .map(key => ({ key, value: mappedStates[key] }))
        .forEach(({key, value}) => {
            mappedStates[key] = {
                set (update) {
                    updateCb.call(this, { [key]: update })
                },
                get () {
                    return value.call(this)
                }
            }
        })
    return mappedStates
}
```

Uses mapState under the hood for consistency and computed {set,get} to be compatible with `v-model`.

[GIST](https://gist.github.com/leodutra/0cfd71fb13e77fa674cc0ceb18477751)

Shyam-Chen:
I made a simple library.

https://github.com/Vanilla-IceCream/vuex-bound

vladimir-chernenko:
You can try to disable `strict`, https://vuex.vuejs.org/en/strict.html

> In strict mode, whenever Vuex state is mutated outside of mutation handlers, an error will be thrown. 

[...]

PR #1308, opened 2018-06-15
Reverse compatable setters for mapState #1085

wferris722:
In response to issue #1085

This PR adds optional setters to mapState that should be intuitive to people familiar with vue's computed setters and vuex's mapMutations helper function. 

The change can be made without breaking existing usage of the mapState function.

Setters mimic rules for mapMutations (a string argument commits a matching mutation, function arguments are dispatched with a context)

PR includes unit tests for above behaviors including namespaces.

Unchanged Behavior:
```
computed: {
  ...mapState(['count'])
  // -- or --
  ...mapState({
    count: 'count'
  })
  // -- or -- 
  ...mapState({
    count: state => state.count
  })
```

Added Behavior:
```
computed: {
  ...mapState({
    count: {
      get: 'count', //functions identically to ...mapState(['count'])
      set: 'count'  //commit('count', ...args)
    },
    data: {
      get: (state) => state.data,
      set: ({commit}, value) => commit('data', value);
    },
  })
}
```

The two behaviors can be mixed:
```
computed: {
  ...mapState({
    count: 'count',
    data: {
      get: 'data',
      set: 'data'
    }
  })
}
```

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -7,7 +7,8 @@
 export const mapState = normalizeNamespace((namespace, states) => {
   const res = {}
   normalizeMap(states).forEach(({ key, val }) => {
-    res[key] = function mappedState () {
+    const getValue = (typeof val === 'object' && 'get' in val) ? val['get'] : val
+    const get = function mappedState () {
       let state = this.$store.state
       let getters = this.$store.getters
       if (namespace) {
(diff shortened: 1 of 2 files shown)


yarsky-tgz:
as i see from your changes your code supports only mutations commits. 

Why `dispatch` is not provided to setters?

wferris722:
I provide a context object, which can be destructured to provide dispatch, commit, getters
So if you wanted to ```dispatch``` (relative to the provided namespace if any), you could use the following:
```
...mapState({
  count: {
    get: 'count',
    set: ({ dispatch }, value) => dispatch('some-action', value)
  }
)
```

santiagofs:
I was about to suggest a similar behavior. I'd like this to be part of Vuex :)

PR #1319, opened 2018-06-23
modify shopping-cart modules cart.js

zjfeng1223:
Code reduction


ktsn:
Thanks for the contribution. However, the previous one would be better since it is more generic.

PR #1322, opened 2018-06-27
feat: add type annotation for the context of actions and mutations

sue71:
The handlers for actions and mutations are bound to the store, but there is no type annotation for this behavior. 

This PR will enable us to use `this` as Store in the handlers.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -28,7 +28,7 @@ export declare class Store<S> {
 
   hotUpdate(options: {
     actions?: ActionTree<S, S>;
-    mutations?: MutationTree<S, S>;
+    mutations?: MutationTree<S>;
     getters?: GetterTree<S, S>;
     modules?: ModuleTree<S>;
   }): void;
@@ -48,7 +48,6 @@ export interface Commit {
 
 export interface ActionContext<S, R> {
   dispatch: Dispatch;
-
   commit: Commit;
   state: S;
   getters: any;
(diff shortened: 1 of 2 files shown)


1000ch:
ping @ktsn

types/index.d.ts
```diff
@@ -86,15 +86,15 @@ export interface StoreOptions<S> {
...
+export type ActionHandler<S, R> = (this: Store<S>, injectee: ActionContext<S, R>, payload: any) => any;
 export interface ActionObject<S, R> {
   root?: boolean;
   handler: ActionHandler<S, R>;
 }
 
 export type Getter<S, R> = (state: S, getters: any, rootState: R, rootGetters: any) => any;
 export type Action<S, R> = ActionHandler<S, R> | ActionObject<S, R>;
-export type Mutation<S> = (state: S, payload: any) => any;
+export type Mutation<S> = (this: Store<S>, state: S, payload: any) => any;
```
ktsn: What is the use case of `this` in mutations? 
I think we should not use that in mutations while it might be useful in actions.

sue71:
@ktsn 
> What is the use case of this in mutations? I think we should not use that in mutations while it might be useful in actions.

I think so too. I just followed the implementation.

ktsn:
@sue71 If there are no use case for mutations, can we remove `this` type for mutations?
Originally, `this` in both the actions and mutations are something like private api to solve an issue on Nuxt. I don't think they are exposed on the user land. But I think it's fine for actions since it's valid use case and well-known for the users now.

sue71:
@ktsn make sense. I'll do so.

PR #1358, opened 2018-07-31
Prepend devtool handler

simplesmiler:
This change adds an API for prepending handlers with `store.subscribe(fn, { prepend: true })` and changes devtool subscriber to use this API.

This fixes https://github.com/vuejs/vue-devtools/issues/678


kiaking:
@ktsn I think we should merge this one...? I could take over and resolve the conflict. Is it still useful do you think?

PR #1365, opened 2018-08-03
Warn about conflicts between state and module

### LINKED ISSUE #1363: Display warning if root state property has the same name with module
If there is a root state property with the same name of a module, the module is prioritized. Right now there is no warning.

## Example:

```
// store
modules: {
    users
},
state: {
    users: []
}
```
`store.state.users` points to the users module and `users` state property is shallowed

There should be a warning pointing out the mistake and intriguing the user to rename the state property.

simplesmiler:
Fixes #1363.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -282,6 +282,13 @@ function installModule (store, rootState, path, module, hot) {
     const parentState = getNestedState(rootState, path.slice(0, -1))
     const moduleName = path[path.length - 1]
     store._withCommit(() => {
+      if (process.env.NODE_ENV !== 'production') {
+        if (moduleName in parentState) {
+          console.warn(
+            `[vuex] state field "${moduleName}" was overridden by a module with the same name`
+          )
+        }
+      }
       Vue.set(parentState, moduleName, module.state)
     })
   }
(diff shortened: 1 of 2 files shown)


src/store.js
```diff
@@ -282,6 +282,13 @@ function installModule (store, rootState, path, module, hot) {
     const parentState = getNestedState(rootState, path.slice(0, -1))
     const moduleName = path[path.length - 1]
     store._withCommit(() => {
+      if (process.env.NODE_ENV !== 'production') {
+        if (moduleName in parentState) {
+          console.warn(
+            `[vuex] state field "${moduleName}" was overridden by a module with the same name`
```
hootlex: @simplesmiler what do u think about including the module name in the latter part as well? `[vuex] state field "${moduleName}" was overridden by a module with the same name ("${moduleName}")`

winniehell:
@simplesmiler do you intend to address the review comment above / complete this pull request?

winniehell:
@ktsn Given that you already approved, do you think this can be merged as is it? Or would you prefer applying the review comment above?

winniehell:
@yyx990803 could you please take a look if this can be merged?

winniehell:
@ktsn @yyx990803 Could you please merge this or comment what's missing? It has been approved over a year ago. :worried:

lavyun:
How can I close this warn, suddenly printed a lot of such warnings in my project., Although it doesn't matter, but the sudden warning was terrible for me.

PR #1369, opened 2018-08-07
fix: Add key to v-for

keng0o:
Updates:
---
- Add key to v-for


examples/shopping-cart/components/ShoppingCart.vue
```diff
@@ -3,7 +3,9 @@
     <h2>Your Cart</h2>
     <p v-show="!products.length"><i>Please add some products to cart.</i></p>
     <ul>
-      <li v-for="product in products">
+      <li
+        v-for="product in products"
+        :key="product.id">
```
zlv2s: Hi, I'm new to Vue, here the product seems to have no id property, i'm not sure if i'm correct or wrong.

PR #1264, opened 2018-05-11
chore: migrate to CircleCI 2 and add Node.js 10

DanielRuf:

PR #1372, opened 2018-08-12
ci: migration to CircleCI 2

gcoguiec:
Since CircleCI 1 will be retired **August 31th**, I've done a couple of things (based on #1264 initial work) to achieve at least a first version of this migration.

#### Done

- [x] As DRY as possible.
- [x] Test-reporting when no additional code was needed.
- [x] Parallel running.
- [x] Build-matrix against multiple node versions (currently `8:latest` and `10:latest`).
- [x] Drops old `circle.yml` file.

#### To be done (or not)

- Test reporting for jasmine tasks (?)
- Coverage [reporting](https://github.com/vuejs/vue/blob/dev/.circleci/config.yml#L37) (?)

#### Additional Notes

I've upgraded `eslint` and `eslint-plugin-vue-libs` to [Vue versions](https://github.com/vuejs/vue/blob/dev/package.json#L94) (junit reporter was slightly broken in previous `eslint` version) and some files were fixed in the process.


ktsn:
Do we really need to run on both Node 8 and 10? Since Vuex is a library for frontend, I think it would be enough to run on one version of Node.

gcoguiec:
Of course! I've updated the configuration accordingly.

DanielRuf:
Well, Node.js 10 includes (breaking) changes so testing both does make much sense (the build environment is still important).

DanielRuf:
To be clear, CI is not just about testing the released product but also testing if the complete toolchain will work (please don't forget the contributors who use Node.js 10 - otherwise it may happen that incompatible dependencies could produce a broken build or cause unforeseen bugs - and there are many - especially well known tools like Grunt and others).

ktsn:
I see where you are coming from. To be honest, it however is too much for Vuex for me. 

Of cause it would be better that we can test all possible cases but it would make ci testing slower and complicated, and may suffer our DX.

On the other point of view, I guess there would not be so much cases that changing Node.js version breaks Vuex workflow as we use it only for tooling and testing. Also it would not be so critical. I'm not sure that we should add extra workflow to test such small risk with the cost.

PR #1373, opened 2018-08-12
Improve strong-typing of getters

evgenyt1:
See small diff. At least in TS3 this improves makes `getters` strongly typed (still allowing generic getters calls).


ktsn:
Passing entire store option type as a type parameter would be too broad. IMO, we should make getter type itself as a type parameter.

kiaking:
Closing the PR due to inactivity. Please open another PR if needed 🤝

Issue #1386, opened 2018-08-30
Confused namespace for get module


leecade:
### Version
3.0.1

### Reproduction link
[https://codesandbox.io/s/vue](https://codesandbox.io/s/vue)

### Steps to reproduce
1. create files in store folder

`store/foo/index.js`

`store/foo/a.js`

2. get module by namespace

```js
computed: {
   ...mapState('foo/a', []), // correct
   ...mapState('foo', []), // correct
   //  ...mapState('foo/index', []), // not works here
}
```

### What is expected?
it make sense if `mapState('foo/index', [])` works as `mapState('foo', [])`

### What is actually happening?
`mapState('foo/index', [])` not work

ktsn:
I don't think this is reasonable. Namespace is not a directory...
If we resolve a module named `index` implicitly, we cannot access its parent module (`foo` module on your example).

PR #1387, opened 2018-08-30
fix: confused namespace for get module

leecade:
see #1386


ktsn:
See https://github.com/vuejs/vuex/issues/1386#issuecomment-417190396

PR #1394, opened 2018-08-31
修复d.ts不能支持tsx的bug

yangdan8:


ktsn:
If you make `$store` optional, you need to check its appearance (i.e. null check) every time you use it...
It's intentional the current declaration because when we use Vuex, `$store` is there in the most of cases.

Issue #875, opened 2017-07-20
Option to not connect to dev tools


Austio:
### What problem does this feature solve?
Hello, I would like to create a pull request that would allow an option to be passed to Vuex constructor to have it not connect to dev tools.  Useful in production so it doesn't connect to __VUE_DEVTOOLS_GLOBAL_HOOK__

This would reduce ease of someone being able to access the store from client.

### What does the proposed API look like?
Proposed API would be something like

```
const store = new Vuex.Store({ 
  connectToDevTools: Bool 
})
```

By default I would default connectToDevTools to true.

posva:
devtools are already hidden in production mode. Why would you need something like this?

Austio:
Hmmm, totally possible i'm missing something in my config, but it looks like in production Vue is not connecting but the store is to window.__VUE_DEVTOOLS_GLOBAL_HOOK__

Notice the Vue is null but store has a reference

![image](https://user-images.githubusercontent.com/1709860/28423509-dd617e98-6d30-11e7-9a13-6f185ce82375.png)

I checked a few other sites that are using vue (like vuejs jobs) and they have this hook.  So long as that thing is present I think the boot will always add it. 

Relevent Plugin: https://github.com/vuejs/vuex/blob/2a67103a1f5fc1448a694e7f83a5f0c6d6bc8262/src/plugins/devtool.js

Line that loads it:
https://github.com/vuejs/vuex/blob/67da6225552e46266ed059c7f0d0128294cd08ed/src/store.js#L61

LinusBorg:
Hm yeah, we could probably not load the plugin if `Vue.config.devtools = false`.

However since the devtools won't show during productiuon anyway, that's more for possible performance-implication than anything else...

ktsn:
Well, I'm not sure why we need to have this config in Vuex side as well. I think there are only cases that it's both enabled (in development) or both disabled (in production).

If you mean that you would like to avoid devtool initialization, it would be better to use `Vue.config.devtools` rather than a new option.

LinusBorg:
@ktsn Yeah, but the point is that currently, no matter what `Vue.config.devtools` is set to, vuex will emit `vuex:init` to the devtools.

If that doesn't do anything in devtools because the devtools didn't receive an initial `init` event, then we don't have an issue. If `vuex:init` does do something, we should simply do:

```javascript
import Vue from 'vue'
if (Vue.config.devtools) {
  // apply plugins
[...]

PR #1407, opened 2018-09-25
Add ability to turn off devtools on vuex by passing an off options

Austio:
This PR adds the ability to turn off devtools on a per vuex instance level.  This is useful for cases where you have multiple vuex stores on a single page.  A good example is a scenario where teams maintain their own portions of a single app that are all loaded, some of which will have an independent vuex store.

With this option, on the dev side i could turn this off per instance so that I can debug a single store. 

By default, this mirrors what is already in vuex, it just allows for turning off when specified.

```
  devtools: false
})
```

Follow through on https://github.com/vuejs/vuex/issues/875

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -63,7 +63,8 @@ export class Store {
     // apply plugins
     plugins.forEach(plugin => plugin(this))
 
-    if (Vue.config.devtools && options.devtools !== false) {
+    const useDevtools = options.devtools !== undefined ? options.devtools : Vue.config.devtools
+    if (useDevtools) {
       devtoolPlugin(this)
     }
   }
--- a/src/store.js
+++ b/src/store.js
@@ -63,7 +63,7 @@ export class Store {
     // apply plugins
     plugins.forEach(plugin => plugin(this))
 
-    if (Vue.config.devtools) {
+    if (Vue.config.devtools && options.devtools !== false) {
       devtoolPlugin(this)
     }
   }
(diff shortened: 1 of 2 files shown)


src/store.js
```diff
@@ -63,7 +63,7 @@ export class Store {
     // apply plugins
     plugins.forEach(plugin => plugin(this))
 
-    if (Vue.config.devtools) {
+    if (Vue.config.devtools && options.devtools !== false) {
```
ktsn: I guess it may be cleaner to use `Vue.config.devtools` as **default value** of `options.devtools` rather than conjunction of both. like:

```js
const devtools = options.devtools !== undefined ? options.devtools : Vue.config.devtools

// ...

if (devtools) {
  // ...
}
```

Austio: I had thought about this approach as well.  I ended up on this way because i think we want to respect someone turning it off in Vue, which acts more like a parent here.

I could see both side of it though, what do you think from that perspective?

Austio: Ah, so if someone has this set to true in dev but ships to prod.  It would still attach to the vue devtools if __VUE_DEVTOOLS_GLOBAL_HOOK__ exists (have saw this exist before when vue devtools are off.)

Going to switch around to your recommendation.

Austio: @ktsn updated :arrow_up:

ktsn:
Thanks!

PR #1425, opened 2018-10-16
Fix false negative when checking `payload: false`

roschaefer:
Here is my example code:
```javascript
    test('commits setContributionId', (done) => {
      const state = {
        comments: []
      }
      testAction(action, 42, state, [
        // ...
        { type: 'isLoading', payload: false}
      ], done)
    })
```
I'm passing `42` as an argument to `testAction` and I want to verify that `commit('isLoading', payload)` is executed. Here the payload is not the payload of the testAction helper (in my case 42) but the second argument of `commit`. That might be `false` and in my case it is. If I change the payload of the last expected commit to be `true` no assertion error is raised. The fix is to make the if condition in the mocked commit method more specific.

### DIFF
--- a/docs/guide/testing.md
+++ b/docs/guide/testing.md
@@ -95,9 +95,7 @@ const testAction = (action, payload, state, expectedMutations, done) => {
 
     try {
       expect(type).to.equal(mutation.type)
-      if (mutation.payload !== undefined) {
-        expect(payload).to.deep.equal(mutation.payload)
-      }
+      expect(payload).to.deep.equal(mutation.payload)
     } catch (error) {
       done(error)
     }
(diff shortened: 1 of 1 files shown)


roschaefer:
I've added one more commit. From the commit message:

# Given a mutation payload, don't skip assertion 

If a user specifies an expected payload, the assertion **must** be
executed.

Let's say your store action looks like this:

```js
const somePayload = undefined
// this is the bug to be exposed, `somePayload` should not be undefined

commit('isLoading', somePayload)
```

And the developer defines an expected payload in the test:

```js
testAction(action, 42, state, [
  //...
  { type: 'isLoading', payload: false }
], done)
```

The test will unexpectedly pass as the assertion is skipped.
Instead we always want to check the assertion if any payload was
defined by the developer.

ktsn:
Maybe we don't need this `if` statement?

kiaking:
Yeah I think we can just remove `if` and be done with it. Always asserting the payload would be much more intuitive I think.

@roschaefer Do you think you'll be able to adjust the change?

kiaking:
@roschaefer Thanks a lot!

PR #1426, opened 2018-10-17
fix: function 'unregister', someplace maybe inaccurate 

zwwill:
Function `parent.getChild(key)` could return `undefined`

### DIFF
--- a/src/module/module-collection.js
+++ b/src/module/module-collection.js
@@ -49,7 +49,7 @@ export default class ModuleCollection {
   unregister (path) {
     const parent = this.get(path.slice(0, -1))
     const key = path[path.length - 1]
-    if (!parent.getChild(key).runtime) return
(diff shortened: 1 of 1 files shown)


src/module/module-collection.js
```diff
@@ -49,7 +49,7 @@ export default class ModuleCollection {
   unregister (path) {
     const parent = this.get(path.slice(0, -1))
     const key = path[path.length - 1]
-    if (!parent.getChild(key).runtime) return
+    if (!parent.getChild(key) || !parent.getChild(key).runtime) return
```
ktsn: I prefer assertion rather than silently ignoring when the specified module does not exist. Can we change the implementation to that?

zwwill: I agree, maybe you can do it better latter

kiaking:
@zwwill Sorry for mentioning in old PR, though do you think it would be possible to include the suggestion from @ktsn to this PR?

PR #1438, opened 2018-11-02
Prevent invalidating subscription iterator 

cngu:
This change shallow copies `_subscribers` before iterating over it, in-case one of the subscription listeners attempts to unsubscribe synchronously and invalidate the iterator.


src/store.js
```diff
@@ -100,7 +100,7 @@ export class Store {
         handler(payload)
       })
     })
-    this._subscribers.forEach(sub => sub(mutation, this.state))
+    this._subscribers.slice().forEach(sub => sub(mutation, this.state))
```
cngu: I'm open to cleaner/faster ways to generate a shallow copy.

cngu:
It can be useful in scenarios where you want to teardown a subscription when a particular mutation occurs (e.g. logout), or mutations that are guaranteed to only be committed N times during the lifetime of the app (e.g. 1 login from OAuth flow, as another login would have to reload the page).

src/store.js
ktsn: I guess it's ok to use `slice` for that.

src/store.js
```diff
@@ -100,7 +100,7 @@ export class Store {
         handler(payload)
       })
     })
-    this._subscribers.forEach(sub => sub(mutation, this.state))
```
ktsn: Could you write some comment to share the context?

ktsn: I think we also need to do the same thing for `_actionSubscribers` in `dispatch` method.
https://github.com/vuejs/vuex/blob/dev/src/store.js#L132

cngu: Done!

cngu: Good catch, I've shared it through a new `notifySubscribers` function.

cngu:
Hi @ktsn, are there any other changes you'd like to see in this PR?

PR #1442, opened 2018-11-06
Eliminate extra word

dcrall:
"You can ~do~ achieve this" -> "You can achieve this"


ktsn:
Thanks for making the PR. However, `do + verb` looks valid English. It is used for emphasis.
https://www.usingenglish.com/forum/threads/166197-I-do-verb

dcrall:
The example you sited is **I do** + verb, which is primarily a conversational construct. This edit does not match that pattern, and I can assure you it does not read as correct English. It is awkward, redundant, and adds no clarity to the statement

dcrall:
Plus, in this sentence, you have three verbs: can, do, and achieve. That is another reason it doesn't match the pattern in the link.

ktsn:
I see. Thank you for clarifying.

PR #1453, opened 2018-11-20
docs: replaced video preview code to a VideoPreview component

Jinjiang:
... for better maintenance and translation.
Ref: https://v0.vuepress.vuejs.org/guide/using-vue.html#using-components
Thanks.


ktsn:
Cool! 👍 

One thing I found is the video is no longer auto played after clicking placeholder image. We also need to click the play button in the iframe video...
This probably is because it is needed to play it in the user's click callback on Chrome while the iframe is now appeared by `v-if` directive.
Can you modify it to behave as same as before?

Jinjiang:
@ktsn well I found it worked in Safari which is my preference. But yes it failed in Chrome. 😅
Sorry for my lack of testing, will check it out soon. 👌
Thanks.

PR #1468, opened 2018-12-08
docs: add a description of providing store to Vue instance

NataliaTepluhina:
_Note:_ This change was already introduced in #858 and it's still crucial for Vuex docs but since this PR exists for 1.5 years and has conflicts with a base branch, I'm opening a new one.

Added a short description of how to provide a store to Vue instance to have access to it from inside Vue components.

**Update:** found docs on injecting store in the `State` doc. I still think they should be moved to `Getting started` part because it's just easier not to miss it in the entry doc.

### DIFF
--- a/docs/guide/README.md
+++ b/docs/guide/README.md
@@ -32,7 +32,7 @@ const store = new Vuex.Store({
 })
 ```
 
-In order to have an access to `$store` property in your Vue components, you need to provide the created store to Vue instance:
+In order to have an access to `this.$store` property in your Vue components, you need to provide the created store to Vue instance. Vuex has a mechanism to "inject" the store into all child components from the root component with the `store` option
 
 ```js
 new Vue({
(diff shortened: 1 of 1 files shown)


kiaking:
@NataliaTepluhina I like this change, and I think we should merging this. I think we might want mention that it's crucial to inject the store to the Vue instance to support SSR as well, though that can be done in following PR I think.

Do you think you can resolve the conflict? 🙏

NataliaTepluhina:
@kiaking sure, let me try :)

kiaking:
Thanks a bunch 🙌

NataliaTepluhina:
@kiaking done! I rephrased a description slightly to make the reader's flow easier and also I decided to leave the description in `state.md` as it was: it will give us a small repeating but nothing critical. Please let me know what do you think!

kiaking:
Thanks a lot! I love it. I think the repeat in `state.md` is sounds reasonable too. If we get any feedbacks that repeating causing some confusion, I think we can deal with that at that time.

Wonderful 🎉

PR #1477, opened 2019-01-01
docs & examples: rewrite state object to functions everywhere

tombo1818:
## Overview
Rewrite state, object to function everywhere.

## Details
To avoid shared state on the server side, we should change our state from object to functions everywhere.

## Related Pages
[stack overflow](https://stackoverflow.com/questions/49557177/vuex-state-returned-as-function-or-object-literal)

If it is wrong, please close this pr:bow:

### DIFF
--- a/docs/guide/README.md
+++ b/docs/guide/README.md
@@ -16,9 +16,9 @@ After [installing](../installation.md) Vuex, let's create a store. It is pretty 
 // Make sure to call Vue.use(Vuex) first if using a module system
 
 const store = new Vuex.Store({
-  state: {
+  state: () => ({
     count: 0
-  },
+  }),
   mutations: {
     increment (state) {
       state.count++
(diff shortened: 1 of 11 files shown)


ktsn:
I'm not sure if we should change the usage examples to the function form while we already have module reuse section https://vuex.vuejs.org/guide/modules.html#module-reuse. Replacing the example does not seem to let the users get the rationale behind it. 

If the module reuse section is not clear for you, I guess improving the section (or move it to more appropriate place like state chapter) would be better.

kiaking:
Yeah, I personally like using function in modules, though not sure if we should do that in static store option as well. Since it's same as Vue Component's `data` property.

@tombo1818 Do you think you can revert back all the static state definition part to be plain object...? (and keep modules as is

kiaking:
@ktsn Do you think it would make sense to update all module state declaration to be a function? Since now, Vue is also ditching the plain object usage, it might make sense to have examples in function in module. If yes, I can make PR for it.

ktsn:
That makes sense to me.

kiaking:
Great. Let's take over this.

PR #1478, opened 2019-01-01
Fix: add `devtools` to StoreOptions interface

haversnail:
Updates `StoreOptions` type to include the [`devtools`](https://vuex.vuejs.org/api/#devtools) option, resolving TypeScript error when option is included in store constructor.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -86,6 +86,7 @@ export interface StoreOptions<S> {
   modules?: ModuleTree<S>;
   plugins?: Plugin<S>[];
   strict?: boolean;
+  devtools?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


Midnighter:
Coming with an outside perspective: Someone creates a type annotation for you which allows using a publicly documented option on the store with typescript. 5 months later this is still not possible. I fail to see how a test here (whatever that should look like) can trump bringing this into the project.

kiaking:
@ktsn We still want this right? If so, I can take this over.

ktsn:
Added a test case and merged in https://github.com/vuejs/vuex/commit/38c11dcbaea7d7e661a1623cabb5aef7c6e47ba7
Thank you!

PR #1487, opened 2019-01-17
docs: added wording for rootState access in module mutations

rabbitwerks:
Added content for accessing Store rootState in module mutations by passing through action. included example code snippets that are well explained and brief. Using this pattern in my current project.

### DIFF
--- a/docs/guide/modules.md
+++ b/docs/guide/modules.md
@@ -66,6 +66,38 @@ const moduleA = {
 }
 ```
 
+More so, some cases may require access to the `rootState` when committing mutations in modules that affect the Vuex Store root state. This can be done easily by passing through `context.rootState` or destructured as `{ rootState }`.
+
+``` js
+Vuex.Store = ({
+  state: {
+    rootSum: 0,
+  }
+})
+
+const moduleA = {
+  //...
+  mutations: {
+    increaseRootSumByValue(state, { rootState, payload: { value, ... } }) {
+      rootState.rootSum += value
+    },
+    increaseRootSumByOne(state, rootState) {
+      rootState.rootSum++;
+    },
+  };
+  actions: {
+    // payload = { value, ... }
+    increaseIfValue({ commit, rootState }, payload) {
+      if(payload.value) {
+        commit.increaseRootSumByValue({ rootState, payload });
+      } else {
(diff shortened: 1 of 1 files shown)


Sharpie360:
@ktsn that does make sense, however consider this case.

the app I am working on right now essentially revolves around one array of objects that lives in the store.js state object. The majority, actually, all of the Vuex mutations I am using touch this rootState array to update the data inside of it. The thing is that i have different Modules writen for the various functionality groups of this app, that i like to keep seperate and organized from each other, while all still mutating the rootState array that the app revolves around.

this is a side-case for sure, and it could be that there is a way better approach to my situation that i haven't discovered yet. That being said, I propose a section in the docs specifically for edge cases like this, and with that, "recommended" ways to handle these edge cases.

Thanks in advance! :smiley:

ktsn:
Sorry, I still do not get why you are writing module mutations which mutates root state...
If it is just for readability (separate logics into small groups), you do not have to split them as modules. You can just declare them in separated files and merge them into one root module.

Sharpie360:
That's pretty much what I've done, as pieced together from not having detailed documentation on situations like this. 

Is there another way to seperate logic without using modules that I can still target one root state data item?

If so, clear documentation of the "right" way to do stuff is needed. Otherwise these situations will become more abundant as time goes on. 

This way my instinct for use of Vuex Modules and someone with less experience will probably follow suit in methodology.

Github.com/Sharpie360/flexbox-util-playground.git

ktsn:
You can use spread operator, `Object.assign` or some utility function to merge your module assets.

```js
// foo-mutations.js
const fooMutations = {
  // ...
}

// bar-mutations.js
const barMutations = {
  // ...
}

// Merge your mutation groups into root module
  state: { /* ... */ },

  mutations: {
    ...fooMutations,
    ...barMutations
  }
})
```

Sharpie360:
That makes sense. I will try this method out, although shown in the project repo, each module contains their own getters, mutations, and actions.

Will the suggested method work in targeting specific exports of the modules to their respective vuex objects?

kiaking:
Closing the PR since it's technically possible to do this, I don't think it's something we should encourage users to do.

PR #1497, opened 2019-01-30
update ja docs

kazupon:
/ping @ktsn  🙏


kazupon:
Thank you for your reviewing!

PR #1508, opened 2019-02-14
Fixed issue#1507 : Memory leak happening while using registerModule/u…

### LINKED ISSUE #1507: Memory leak happening while using registerModule/unregisterModule.
### Version
3.1.0

### Reproduction link
[https://jsfiddle.net/mrj8spu0/](https://jsfiddle.net/mrj8spu0/)

### Steps to reproduce

1. Open the JSFiddle link mentioned as part of this issue.
2. Check and record the heap memory in the browser on initial stage.
3. Click on the "Register" button to dynamically register Module B 100 times
4. Now we will see an increase in heap memory on account of registering module 100 times.
5. Click on "Un-Register" button to dynamically un-register Module B 100 times
6. Check the heap memory again, at this point will see an jump in memory usage on account of un-registering 100 module. 

### What is expected?
Heap memory generated on step 3 & 5 must be cleared after step 6. (Final stage)

Heap memory of the browser is not getting cleared.

---
We are observing memory leak of Store._vm while using registerModule/unregisterModule extensively.
On heap profiling we found that the oldVm instance is not getting garbage collected.

Probable Solution:
De-reference the oldVm instance on the same function scope where its getting instantiated (i.e. resetStoreVM() )

The above can be achieved by adding following code inside resetStoreVM function.

oldVm.$destroy(); 
oldVm = null; // This is the missing code

Attached google chrome (V 72.0.3626.96) heap snapshot and file.
heapsnapshot

[Heap-20190213T220212.heaptimeline.zip](https://github.com/vuejs/vuex/files/2861246/Heap-20190213T220212.heaptimeline.zip)

parth67:
Fixes #1507 

Closure function was fixed with helper function partial which doesn't retain oldVm in environment.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -291,7 +291,7 @@ function resetStoreVM (store, state, hot) {
         oldVm._data.$$state = null
       })
     }
-    Vue.nextTick(() => { oldVm.$destroy(); oldVm.computed = null })
+    Vue.nextTick(() => oldVm.$destroy())
   }
 }
 
--- a/src/store.js
+++ b/src/store.js
@@ -1,7 +1,7 @@
 import applyMixin from './mixin'
 import devtoolPlugin from './plugins/devtool'
 import ModuleCollection from './module/module-collection'
-import { forEachValue, isObject, isPromise, assert } from './util'
+import { forEachValue, isObject, isPromise, assert, partial } from './util'
 
 let Vue // bind on install
 
@@ -256,7 +256,7 @@ function resetStoreVM (store, state, hot) {
   const computed = {}
   forEachValue(wrappedGetters, (fn, key) => {
     // use computed to leverage its lazy-caching mechanism
-    computed[key] = () => fn(store)
+    computed[key] = partial(fn, store)
     Object.defineProperty(store.getters, key, {
       get: () => store._vm[key],
       enumerable: true // for local getters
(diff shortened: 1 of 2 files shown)


src/store.js
```diff
@@ -256,7 +256,7 @@ function resetStoreVM (store, state, hot) {
   const computed = {}
   forEachValue(wrappedGetters, (fn, key) => {
     // use computed to leverage its lazy-caching mechanism
-    computed[key] = () => fn(store)
+    computed[key] = partial(fn, store)
```
ktsn: Adding some comment to share the context of this memory leak would be nice.

src/util.js
```diff
@@ -64,3 +64,9 @@ export function isPromise (val) {
 export function assert (condition, msg) {
   if (!condition) throw new Error(`[vuex] ${msg}`)
 }
+
+export function partial (fn, ...bindargs) {
```
ktsn: Let's simplify to:

```js
export function partial (fn, arg) {
  return function () {
    return fn(arg)
  }
}
```

...because spread parameter bloats the output code.

parth67: simplified to have single argument.

src/store.js
parth67: Added comment on code

src/store.js
```diff
@@ -289,7 +291,7 @@ function resetStoreVM (store, state, hot) {
         oldVm._data.$$state = null
       })
     }
-    Vue.nextTick(() => oldVm.$destroy())
+    Vue.nextTick(() => { oldVm.$destroy(); oldVm.computed = null })
```
ktsn: Why do you assign `null` to `oldVm.computed`?

parth67: This was accidental and absolutely not required for this fix. Corrected.

eladcandroid:
Can you please merge it?

PR #1510, opened 2019-02-17
Add support for namespace functions in mapXXX helpers

brophdawg11:
This is a stab at adding the functionality requested by https://github.com/vuejs/vuex/issues/863.

I also noticed that there was another attempt at this in https://github.com/vuejs/vuex/pull/924, but the diff seems to be beyond the scope of just the change.

Unit tests have been added, but no changes have yet been made to TypeScript types or documentation, etc.

### DIFF
--- a/src/helpers.js
+++ b/src/helpers.js
@@ -11,7 +11,13 @@ export const mapState = normalizeNamespace((namespace, states) => {
       let state = this.$store.state
       let getters = this.$store.getters
       if (namespace) {
-        const module = getModuleByNamespace(this.$store, 'mapState', namespace)
+        if (typeof namespace === 'function') {
+          namespace = namespace.call(this, this)
+          if (namespace.charAt(namespace.length - 1) !== '/') {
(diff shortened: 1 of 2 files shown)


src/helpers.js
```diff
@@ -11,7 +11,13 @@ export const mapState = normalizeNamespace((namespace, states) => {
       let state = this.$store.state
       let getters = this.$store.getters
       if (namespace) {
-        const module = getModuleByNamespace(this.$store, 'mapState', namespace)
+        if (typeof namespace === 'function') {
+          namespace = namespace.call(this, this)
```
hufeng: Because the component may be loaded multiple times, the namespace of the subsequent component is always the result of the first execution, so the code cannot override namespace = namespace.call(this, this), we can use the local variable let ns = namespace.call(this, this)

brophdawg11:
Closing as we have released our internal solution to this as a separate package: https://www.npmjs.com/package/@urbn/vuex-helpers

PR #1512, opened 2019-02-20
Added Scrimba lessons

perborgen:
Here are the Scrimba lessons localized and with an `override.styl` file for the styling. Did I do it right?


perborgen:
Added a margin-top of 16 px, as that's similar to what the paragraphs have.

PR #1517, opened 2019-03-07
Make mutation and action payload optional in definition file

### LINKED ISSUE #1491: Optional attribute declared as required on typescript declaration
### Version
3.1.0

### Reproduction link
[https://github.com/vuejs/vuex/blob/dev/types/index.d.ts](https://github.com/vuejs/vuex/blob/dev/types/index.d.ts)

### Steps to reproduce
Test a mutation without payload on typescript

### What is expected?
Typescript not complain about missing payload

Typescript complains about missing payload

---
change: 

export type Mutation<S\> = (state: S, payload: any) => any;

to:

export type Mutation<S\> = (state: S, payload?: any) => any;

davidsandoz:
When testing a mutation without payload in TypeScript, it complains that about the missing payload, even though it's not mandatory.

This PR simply makes the payload optional.

Fixes #1491

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -97,7 +97,7 @@ export interface StoreOptions<S> {
   strict?: boolean;
 }
 
-export type ActionHandler<S, R> = (this: Store<R>, injectee: ActionContext<S, R>, payload: any) => any;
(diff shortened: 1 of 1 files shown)


davidsandoz:
@ktsn, change for action payload is now done as well.

PR #1530, opened 2019-04-05
Genericity for mutations and actions.

wjw99830:
Add genericity for mutations and actions to hint the commit and dispatch methods.


kiaking:
Thanks for the PR! Though I think this is too broad in terms what it get as a return. It might be nice to have this generics, though it only works on RootState and module `dispatch` and `commit` will not be typed anyway. I think it's better to use `mapXXX` helpers since it kinda gives the same effect.

Issue #1247, opened 2018-04-22
Support ES6 import from <script type="module"> tag inside browser


gorbypark:
### What problem does this feature solve?
Using vuex in the browser is possible, by using a standard html ```<script src='"vuex.js">``` tag, but if you want to use something like mapGetters, it require something like ```import { mapGetters } from 'vuex'```.  This is in theory possible in the browser by using a tag like 
```
<script type="module>
  import { mapGetters } from './vuex.js'
  ...
</script>
``` 
but appears to not work with vuex.  Vue itself, using the vue.esm.browser.js bundle from unpkg works this way, but if vuex.esm.js is used it produces an error.
```
ReferenceError: process is not defined
    at Vue.mappedGetter (vuex.esm.js:848)
    at Watcher.get (vue.js:3140)
    at Watcher.evaluate (vue.js:3247)
    at Proxy.computedGetter (vue.js:3503)
    at Proxy.eval (eval at createFunction (vue.js:10667), <anonymous>:3:144)
    at Vue._render (vue.js:4535)
    at Vue.updateComponent (vue.js:2788)
    at Watcher.get (vue.js:3140)
    at new Watcher (vue.js:3129)
    at mountComponent (vue.js:2795)
```

### What does the proposed API look like?
```import { ... } from 'vuex.esm.js'``` should work inside the browser.

ktsn:
FYI, you can use `mapXXX` helpers via `Vuex.mapState` etc on umd bundle.

pgdata:
I think there's a problem with build process: vuex.esm.js should not refer to "process.env.NODE_ENV", this is why you got _"ReferenceError: process is not defined"_
Remove every reference to "process.env.NODE_ENV" from vuex.esm.js and you could use:

`import Vuex from './vuex.esm.js'`

TimvdLippe:
I have a PR up for a couple of months now in #1401 which seems to be popular in the community. Could any maintainer maybe take a look?

PR #1533, opened 2019-04-11
feat(esm build): build ES modules for browser

zigomir:
Fixes https://github.com/vuejs/vuex/issues/1247 

Adds minified build which is missing in https://github.com/vuejs/vuex/pull/1401

Similar effort as in https://github.com/vuejs/vue-router/pull/2705


Fenzland:
When can we use this?
```javascript
import * as Vuex from 'https://unpkg.com/vuex@3.1.0/dist/vuex.esm.browser.js';
```

zigomir:
@Fenzland I believe you'll get that only after new version of vuex will be released.

PR #1546, opened 2019-05-03
Implementing a cache for the gettersProxy object creation

### LINKED ISSUE #1539: makeLocalGetters performance hit on ssr
### Version
3.1.0

### Reproduction link
[https://codepen.io/frankcs/pen/NmOLNX](https://codepen.io/frankcs/pen/NmOLNX)

### Steps to reproduce
Using vuex on ssr following the ssr guide. I have created an application with a large number of components and modules and therefore a large number of namespaced getters. Every time such a getter is invoked the makeLocalGetters function is called iterating trough all the getters in the store just to create the gettersProxy object. This is overkill for ssr performance and is being flagged as a CPU consuming process in the node profiler. 
I have managed to double the number of requests per second my ssr app can handle just by using memoization on the gettersProxy object creation. I think this is really impacting performance.

### What is expected?
Namespaced getters invokation not to be inefficient with large numbers of getters/modules

A lot of hits to makeLocalGetters making performance go down

frankcs:
Fixes #1539


src/store.js
```diff
@@ -393,27 +393,37 @@ function makeLocalContext (store, namespace, path) {
...
-    // extract local getter type
-    const localType = type.slice(splitPos)
-
-    // Add a port to the getters proxy.
-    // Define as getter property because
-    // we do not want to evaluate the getters in this time.
-    Object.defineProperty(gettersProxy, localType, {
-      get: () => store.getters[type],
-      enumerable: true
+  if (cacheStore !== store) {
```
ktsn: This way of clearing cache is not comprehensive. Getters may be changed via `registerModule`, `unregisterModule` or `hotUpdate`.

frankcs: Thank you so much for your reply! Could you give me a hint here?
I see that all those functions end up calling `resetStoreVM`, do you think that is the proper place to clear the cache? 
That way I could also get rid of that `cacheStore` variable

src/store.js
```diff
@@ -393,27 +395,31 @@ function makeLocalContext (store, namespace, path) {
   return local
 }
 
+let makeLocalGettersCache = {}
```
ktsn: Can we place the cache into store instance so that we can avoid global mutable variable?

src/store.js
```diff
@@ -252,6 +252,8 @@ function resetStoreVM (store, state, hot) {
 
   // bind store public getters
   store.getters = {}
+  // reset local getters cache
+  makeLocalGettersCache = {}
```
ktsn: Let's initialize it with `Object.create(null)` to avoid false positive check with its prototype functions.

afwn90cj93201nixr2e1re:
@ktsn can you check my issue too? https://github.com/vuejs/vue-devtools/issues/1058

superbervin:
image
why not just use 
oldVm =null;
since there is still memory leak problem now, please reconsider this issue ,tks [#1507 ](https://github.com/vuejs/vuex/issues/1507)

PR #1551, opened 2019-05-15
对象展开运算符

dajiaoyisheng:
It is a Stage 4 proposal for ECMAScript.


ktsn:
Thank you for the PR but I've removed the statement about stage to avoid frequent update of docs.

PR #1552, opened 2019-05-15
fix memory leak #1507

wolfgangwalther:
The fix in PR #1508 was not complete, the anonymous function for the getter `() => store._vm[key]` was preserving the previous context and therefore `oldVm` as well.


wolfgangwalther:
resolved in #1546

PR #1563, opened 2019-06-01
Typescript: Use generics for payload/result in commit/dispatch

WIStudent:
I am using Typescript to write my store, like this for example:
```typescript
interface State {
    counter: number;
}

interface AddToCounterPayload {
    amount: number;
}

interface SubstractFromCounterPayload {
    amount: number;
}

type SubstractFromCounterResult = number

const store = new Store<State>({
    mutations: {
        addToCounter: (state, payload: AddToCounterPayload): void => {
            state.counter += payload.amount;
        }
    },
    actions: {
        substractFromCounter: async (context, payload: SubstractFromCounterPayload): Promise<SubstractFromCounterResult> => {
            const addToCounterPayload: AddToCounterPayload =  {amount: -1 * payload.amount};
            context.commit('addToCounter', addToCounterPayload)
            return context.state.counter;
        }
    },
    state: {
        counter: 0
    }
})
```

Triggering a mutation like this will currently not check if the payload has the correct type
```typescript
// correct payload type
store.commit('addToCounter', {amount: 1});
// incorrect payload type
store.commit('addToCounter', {amount: '1'});
// Current solution to ensure correct type
const payload: AddToCounterPayload = {
  amount: 1
}
store.commit('addToCounter', payload);
```
But there are cases where I would prefer to build the payload inline, that's why I would like to propose using generics to set the payload type:
```typescript
store.commit<AddToCounterPayload>('addToCounter', {amount: 1});
```

Triggering actions would work similary but additionaly the type of the returned value can be set using the second generic parameter
```typescript
const result = store.dispatch<SubstractFromCounterPayload, SubstractFromCounterResult>('substractFromCounter', {amount: 1}); 
// type of result will be Promise<SubstractFromCounterResult>
```
By setting the default values of the generics to `any` the behavior of `commit` and `dispatch` won't change if called without generics.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -39,12 +39,12 @@ export declare class Store<S> {
 export declare function install(Vue: typeof _Vue): void;
 
 export interface Dispatch {
-  (type: string, payload?: any, options?: DispatchOptions): Promise<any>;
-  <P extends Payload>(payloadWithType: P, options?: DispatchOptions): Promise<any>;
+  <P = any, R = any>(type: string, payload?: P, options?: DispatchOptions): Promise<R>;
+  <P extends Payload, R = any>(payloadWithType: P, options?: DispatchOptions): Promise<R>;
 }
 
 export interface Commit {
-  (type: string, payload?: any, options?: CommitOptions): void;
+  <P = any>(type: string, payload?: P, options?: CommitOptions): void;
   <P extends Payload>(payloadWithType: P, options?: CommitOptions): void;
 }
 
--- a/types/test/index.ts
+++ b/types/test/index.ts
@@ -15,12 +15,15 @@ namespace StoreInstance {
   store.getters.foo;
 
   store.dispatch("foo", { amount: 1 }).then(() => {});
+  store.dispatch<{amount: number}>("foo", { amount: 1 }).then(() => {})
+  store.dispatch<{amount: number}, string>("foo", { amount: 1 }).then((value) => {})
   store.dispatch({
     type: "foo",
     amount: 1
   }).then(() => {});
 
   store.commit("foo", { amount: 1 });
+  store.commit<{amount: number}>("foo", { amount: 1 })
   store.commit({
     type: "foo",
     amount: 1


lmiller1990:
Even if you have the correct type at the `dispatch`, there is no guarantee that the `payload` matches with the one expected in the action declaration in the store, though. Is this really useful to have one (but not the other)?

The way I do this (before I abandoning Vuex w/ TS) was:

```ts
const payload: IPayload = {
  // ...
}

this.$store.commit('mutation', payload)
```

So you get the same result your proposed changes bring, without changing the library, at least for the commit.

The generics might be useful for the resolved value of a `dispatch`.

Just some input - perhaps I am missing something else, or there is a use case I haven't considered.

WIStudent:
To make sure that the payload types are matching I export them in the store definition and import them where I use commit/dispatch. Using my initial example above, I would write something like this:

definition.ts
```typescript
export interface AddToCounterPayload {
    amount: number;
}

export interface SubstractFromCounterPayload {
    amount: number;
}

export type SubstractFromCounterResult = number
...
```
and in a different file:
```typescript
import {AddToCounterPayload, SubstractFromCounterPayload, SubstractFromCounterResult} from definition.ts

commit<AddToCounterPayload>('addToCounter', {amount: 1})
dispatch<SubstractFromCounterPayload, SubstractFromCounterResult>('substractFromCounter', {amount: 1}); 
```

This has the advantage that when I am making changes to the payload types it will automatically affect the commit/dispatch calls and throw an typescript error if the passed object is no longer compatible to the type.

As for
```typescript
const payload: IPayload = {
  // ...
}

this.$store.commit('mutation', payload)
```
I am aware that this is also an option to ensure the correct payload type. But I have often cases there the content of the payload comes from other functions that return an object that basically just extends the payload type, for example:
```typescript
interface Data {
  amount: number
  price: number
  articleName: string
}

const loadData = (): Data => {
...
}

...
const data = loadData();
commit<AddToCounterPayload>('addToCounter', data)

// Do stuff with the other fields of data
...
```

Using the generic I can pass the whole data object as long as it extends AddToCounterPayload. Personally I think this is more readable/less bloated than creating a temporary object just for the typescript check.

As for compatability, by setting the default values to `any` this doesn't introduce any breaking changes. The commit/dispatch versions there the type is included in the payload actually already support setting the payload type using generics.

kiaking:
I think this is very similar to PR #1530. I'm really not sure why could be useful since it's just passing types at the "same place". For example;

```ts
commit<AddToCounterPayload>('addToCounter', data)
```

The above wouldn't really matter since you know what your committing in both generics, and in function argument. In my opinion, this is really just making sure you don't have typo in payload, but the action name is not typed so... it's really not checking anything... even worth, it might lead to false positive.

I really think we should encourage `mapXXX` helpers for now. I agree that we need better TS support for Vuex, and we should do much better in the next version of Vuex. Though, I don't think this is the way to go about it.

WIStudent:
I am using these generics in a large project mostly in actions that call other actions and mutations. Here is an example how I usualy define an action. 

action.ts
```typescript
import {Commit, Dispatch} from './genericTypings';
import getUser from './getUser';
import {HandleUserActionPayload, HandleUserActionResult} from "./handleUser";
import {SetUserMutationPayload} from './setUserMutation';

interface Context {
  commit: Commit;
  dispatch: Dispatch;
}
export async function action ({commit, dispatch}: Context): Promise<void> {
  const user = await getUser();
  const handledUser = await dispatch<HandleUserActionPayload, HandleUserActionResult>('HANDLE_USER', user);
  commit<SetUserMutationPayload>('SET_USER', handledUser);
}
```

genericTypings.ts
```typescript
import {Commit as VuexCommit, CommitOptions, Dispatch as VuexDispatch, DispatchOptions} from "vuex";

// Make the generics work without having to modify vuex at the moment
export type Commit = VuexCommit & {
  <P = any>(type: string, payload?: P, options?: CommitOptions): void;
}
export type Dispatch = VuexDispatch & {
  <P = any, R = any>(type: string, payload?: P, options?: DispatchOptions): Promise<R>;
}

```

getUser.ts
```typescript
export interface User {
  id: string;
  name: string
}

export default async function getUser (): Promise<User> {
  // Image some library call to get the user
  return {
    id: 'id',
    name: 'name'
  };
}
```
handleUser.ts
```typescript
export interface HandleUserActionPayload {
  id: string;
  name: string;
}

export interface HandleUserActionResult {
  id: string;
  name: string;
  created: string;
}
```
setUserMutation.ts
```typescript
export interface SetUserMutationPayload {
  id: string;
  name: string;
  created: string;
}

```
Before we added the generics we had huge problems whenever someone changed a function (in the example `getUser`), those return value is passed to an action/mutation, but didn't check the whole code base if they needed to adjust actions/mutations. With the generics, we get an typescript error, if the value returned by a function and the required action/mutation payload does not fit. Keep in mind that these actions/mutations are not necessary defined in the same file nor part of the same vuex modules (our project for example has over 60 modules).

WIStudent:
dl;dr However vuex's future typings might look like, they should throw an error whenever commit/dispatch is called with an incompatible payload. This here was just the best solution we came up with within our project without redoing the whole vuex's typings.

kiaking:
Good point yes. Having error thrown might help people a bit more. But I hope we do much better in terms on these kind of typings in next Vuex iteration 💪

rcidaleassumpo:
@WIStudent That's a great solution, I'm trying to get our Commit's  inside of  the module to follow in what's being set in the mutations. Since if I understood correctly, commit's are mutations. So if you do a commit(), it should give you the list of mutations listed in the module. As is now, there isn't much help that you get. Setting the type of the type argument argument to string is not very helpful. 
@kiaking Is there anywhere listing how you guys are dealing with type information in the next Vuex iteration ?

kiaking:
@rcidaleassumpo Not yet at the moment. We are planning to open an RFC after Vue 3 general release 👍

Issue #1571, opened 2019-06-26
Actions mapped to namespaced modules pass unexpected explicit `undefined` payload to `store.dispatch`


markrian:
### Version
3.1.1

### Reproduction link
[https://codepen.io/markrian/pen/vqeGZZ?editors=0010](https://codepen.io/markrian/pen/vqeGZZ?editors=0010)

### Steps to reproduce
1. Open https://codepen.io/markrian/pen/vqeGZZ?editors=0010
1. Open the browser's console
1. Observe the differing logged arguments passed to `store.dispatch` depending on whether it's a root-mapped action or a namespace-mapped action (`[ "foo" ]` and `[ "childModule/foo", undefined ]` respectively - note the extra `undefined` argument in the second case; the different action name is of course expected)

Note that this only really applies in a testing scenario, where the test spies on `store.dispatch` to ensure the component dispatches the expected actions.

### What is expected?
`store.dispatch` receives exactly the same arguments regardless of whether the actions are namespace-mapped or not.

### What is actually happening?
`store.dispatch` receives a payload argument of `undefined` for namespace-mapped actions.

---
I encountered this when refactoring a store into a namespaced module, and found that my expectations needed to change: that `store.dispatch` received an extra `undefined` argument.

I'm not sure if this should be considered a bug in Vuex, or in my testing approach of spying on `store.dispatch` :man_shrugging:

gitlab-winnie:
This is happening because of https://github.com/vuejs/vuex/blob/v3.1.1/src/store.js#L360 (explicitly passing `payload` to dispatch even if `undefined`). A simple (but admittedly not elegant) fix would be

```diff
-  return store.dispatch(type, payload)
+  return payload ? store.dispatch(type, payload) : store.dispatch(type)
```

hugoazevedosoares:
I can work on this issue. Seems preety simple to implement @gitlab-winnie solution.

kiaking:
Closing due to it might already fixed. See comment https://github.com/vuejs/vuex/pull/1676#issuecomment-618341755. Please feel free to open another issue if this is something that we should solve.

PR #1573, opened 2019-07-03
fix(#1571): remove undefined payload from store.dispatch

hugoazevedosoares:
Fixed the issue #1571

Runned all tests and they were ok, except for tsc/types that didn't work on my computer, by that was before the change was made. Also, the change has no reason to break this kind of test.

The solution was proposed by @gitlab-winnie .

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -360,7 +360,7 @@ function makeLocalContext (store, namespace, path) {
         }
       }
 
-      return store.dispatch(type, payload)
+      return payload ? store.dispatch(type, payload) : store.dispatch(type, payload)
     },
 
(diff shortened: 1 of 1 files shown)


Blackskyliner:
The proposal was

```
rreturn payload ? store.dispatch(type, payload) : store.dispatch(type)
```

but you implemented

```
return payload ? store.dispatch(type, payload) : store.dispatch(type, payload)
```

Thus nothing would change with this commit?

kiaking:
Yea, I think the implementation is wrong. @hugoazevedosoares Could you apply what @Blackskyliner mentioned? Also, I think we should have test case for this (as a regression test).

PR #1574, opened 2019-07-08
Refactor: extract a method for duplicate environment judgement

boomler:
There are 17 times of environment judgement in `store.js`.  Extract it to  `util.js`


ktsn:
Thanks for the PR. There are this kind of PRs before but we intentionally write it inline to let uglifyjs strip it in build process.

PR #1586, opened 2019-07-24
.gitignore update to ignore Intellij IDEA and vscode pr…

rummanhasnayeen:
…oject files

.gitignore update to ignore Intellij IDEA project files and vscode project files


ktsn:
Thanks for the PR. However editor specific configuration files should be placed in global gitignore in each local machine otherwise we have to maintain a big list of editor configuration files.

PR #1599, opened 2019-08-07
Update modules.md

SagePacheco:
Minor language correction.


ktsn:
Seems like #1643 would be more appropriate. Thanks for the PR though!

PR #1600, opened 2019-08-08
Fix: " id='app' "

stephentian:


ktsn:
The current code is actually correct. `el: '#app'` will look through the current document tree rather than the template.

PR #1601, opened 2019-08-12
fix(docs): Clearify state object type

### LINKED ISSUE #1587: Storing component instances creates a circular reference loop in strict mode
### Version
3.1.1

### Reproduction link
[https://jsfiddle.net/Sergio_fiddle/zkrsa4g1/3/](https://jsfiddle.net/Sergio_fiddle/zkrsa4g1/3/)

### Steps to reproduce
Store a pointer to a component instance in Vuex with `strict: true`.

### What is expected?
No errors. 

Vuex is adding watchers to all nested properties to warn if there are non-immutable changes inside a object (because of the strict mode help warnings) and does not stop when its entering itself. 

In other words: when entering a component we can re-enter the store as a component property, and in the store we have a component that has a store etc etc ... we are in a loop

---
I think either:
 - Vuex should officially discourage from saving pointers of instances of Vue components in it

**or** 

 - add a check to stop going deeper if the nested object it the Vuex instance itself

SergioCrisostomo:
Clearify in the docs that Vuex state object should be a plain object that follows the same rules as `data` in a Vue instance. This addition will clarify what is "good practise" to use in Vuex.

closes #1587 
closes https://github.com/vuejs/vue/pull/10379

### DIFF
--- a/docs/guide/state.md
+++ b/docs/guide/state.md
@@ -8,6 +8,8 @@ Vuex uses a **single state tree** - that is, this single object contains all you
 
 The single state tree does not conflict with modularity - in later chapters we will discuss how to split your state and mutations into sub modules.
 
+The data you store in Vuex follows the same rules as the `data` in a Vue instance, ie the state object must be plain. **See also:** [Vue#data](https://vuejs.org/v2/api/#data).
+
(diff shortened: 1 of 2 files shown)


ktsn:
Thanks!

PR #1603, opened 2019-08-14
refactor zero padding function

anhmv:


ktsn:
Well, it's not identical with the previous version.

PR #1605, opened 2019-08-14
Add warning when setting state of a non-existent state property name 

### LINKED ISSUE #1120: No warning/error is given when setting state of a non-existent state property name
### Version
3.0.1

### Reproduction link
[https://jsfiddle.net/Sergio_fiddle/m99jmqsa/](https://jsfiddle.net/Sergio_fiddle/m99jmqsa/)

### Steps to reproduce
Open the jsFiddle

### What is expected?
Two possible cases:

a) Vuex would create a property in state and the getter would be loaded with the data

b) a warning/error would be thrown stating "you are setting data to a non-existent property" 

Vuex silently fails

SergioCrisostomo:
This PR add the possibility to be warned in development for the problem described in 1120.
It adds a unit test for that scenario.

While testing this I found out another unit test that needed adjustment and did that in a separate commit. 

Fixes #1120

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -270,9 +270,29 @@ function resetStoreVM (store, state, hot) {
   // some funky global mixins
   const silent = Vue.config.silent
   Vue.config.silent = true
+
+  let $$state = state
+
+  if (process.env.NODE_ENV !== 'production') {
+    $$state = new Proxy($$state, {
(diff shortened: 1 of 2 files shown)


kiaking:
Closing due to the original issue update. See more at https://github.com/vuejs/vuex/issues/1120.

PR #1608, opened 2019-08-15
webpack preset useBuiltIns usage toast promise error

linlinyang:
when we use webpack,
and set babel presets property useBuiltIns is 'usage'
then the polyfill will be loaded when Promise used
so this way has a bug when we webpack,if we use vuex without used Promise,always show the error: vuex requires a Promise polyfill in this browser.

so I have change the validate Promise way to:

```js
try {
  new Promise(() => {})
} catch (error) {
  assert(false, `vuex requires a Promise polyfill in this browser.`)
}
```

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -20,7 +20,7 @@ export class Store {
       // and set babel presets property useBuiltIns is 'usage'
       // then the polyfill will be loaded when Promise used
       // so this way has a bug when we webpack,if we use vuex without used Promise,always show this error
-      //assert(typeof Promise !== 'undefined', `vuex requires a Promise polyfill in this browser.`)
+      // assert(typeof Promise !== 'undefined', `vuex requires a Promise polyfill in this browser.`)
       try {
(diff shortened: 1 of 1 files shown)


ktsn:
But we already have actual promise usages in the code, for example https://github.com/vuejs/vuex/blob/dev/src/store.js#L144.
Is this modification really solve your problem?

linlinyang:
@ktsn But if i just import vuex and didn't new vuex store instance.

linlinyang:
> But we already have actual promise usages in the code, for example https://github.com/vuejs/vuex/blob/dev/src/store.js#L144.
> Is this modification really solve your problem?

But your check promise in the constructor function

ktsn:
I mean, if you need actual promise usage to inject polyfill, it is already satisfied by other part of the code. I suspect this PR actually solve your problem. Can you provide a reproducible case?

kiaking:
Closing due to inactivity.

PR #1609, opened 2019-08-19
replace Array indexof with includes

faithyang:


ktsn:
Thanks but includes does not work on IE.

PR #1611, opened 2019-08-29
docs: Patch preserveState option usage

theneverstill:


kiaking:
Thanks for the PR! I think this is true, but in that case we should fixing the [SSR docs at this section](https://ssr.vuejs.org/guide/data.html#store-code-splitting) as well.

@ktsn Could you double check if its true where we should call `{ preserveState: false }` on server side and `{ preserveState: true }` at the client side?

ktsn:
Yes that's right.

kiaking:
I've opened PR at https://github.com/vuejs/vue-ssr-docs/pull/281. Let's see how it goes.

kiaking:
This has been taking care of at the https://github.com/vuejs/vue-ssr-docs/pull/281#event-3339547032 👍

yooouuri:
How can I set `preserveState` to false on the client and true on the server if I register my components trough the contructor?


```
const stores = {
  ...
}

export function createStore() {
  const modules = {}

  for (const namespace in stores) {
    modules[namespace] = {
      namespaced: true,
      ...stores[namespace],
    }
  }

  return _createStore<State>({
    modules,
    devtools: false,
    plugins: [vuexLocal.plugin],
  })
}
```

PR #1625, opened 2019-10-04
action subscribers would receive a call even when the action failed

fabalint:
That cause might be a promise.rejection or js error.

This way the action subscribers of the 'after' would not miss the end of the action,
as they would receive the existing 'action' and 'state' parameters, and the additional 'isSuccess' and 'errorResult' values. 
This applies for the then-side of the promise too, but there they would receive the result.

The true/false value might be useful for after action handlers, but the result might be too tempting to be used for modification for example, might be that too much info is handled out in this way. The same goes for the success-side of the after subscribers too.

While reviewing mind that: I cannot access github directly (firewalls), so this was a manual insertion of the locally working code. Typos might came along! (: In case there is a problem, tell me and I fix that, or say a thank you for your fix (:

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -148,7 +148,7 @@ export class Store {
       try {
         this._actionSubscribers
           .filter(sub => sub.after)
-          .forEach(sub => sub.after(action, this.state))
+          .forEach(sub => sub.after(action, this.state, /*isError:*/false, res))
       } catch (e) {
         if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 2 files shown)


kiaking:
Thanks for the PR! But I think this can be handle by #1531 where we define dedicated error handling hook instead of handling inside single `after` hook.

Let me followup on that PR and see how it goes.

kiaking:
New error handler is added via https://github.com/vuejs/vuex/pull/1740 👍  You may catch errors there.

PR #1630, opened 2019-10-22
Vuex TypeScript Docs

jefer590:
I noticed that Vuex has already a couple of interfaces to support Typed stores but I didn't found any info about this. So this guide that I made for the Vuex docs is pretty much what i've done in my experience. 

Feedback is very welcome 💪

### DIFF
--- a/docs/guide/typescript.md
+++ b/docs/guide/typescript.md
@@ -1,0 +1,157 @@
+# TypeScript
+
+Vuex supports TypeScript for state, actions, mutations, getters **using the module system**
+
+## Getting Started
+
+To work with typed Vuex, we will need to define a RootState structure first. This will be needed in our modules later.
+
+Define your `RootState` type.
+
+```ts
+export type RootState = {
+  hello: string
+}
+```
+
+And then, use it in your `main.ts` file, which is the entry point file of your Vuex store
+
+```ts
+import Vue from 'vue'
+import Vuex from 'vuex'
+import { RootState } from '~/store/types/RootState'
+
+Vue.use(Vuex)
+
+export default new Vuex.Store<RootState>({
+  state: {
+    hello: 'world' // Typed
+  }
+})
+```
+
+Before creating the module, create a type structure for your module state
+
+```ts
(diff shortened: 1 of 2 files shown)


docs/guide/typescript.md
```diff
@@ -0,0 +1,157 @@
...
+    const { data } = await axios.get('URL')
+    context.commit('mutation', data)
+    await context.dispatch('myotheraction')
+  }
+}
+```
+
+## Mutations
+
+Mutations can be use with the `MutationTree<S>` interface where `S` is the current store type. It works exactly the same as a normal mutation but with the power of types.
```
ederchrono:
```suggestion
Mutations can be used with the `MutationTree<S>` interface where `S` is the current store type. It works exactly the same as a normal mutation but with the power of types.
```

docs/guide/typescript.md
```diff
@@ -0,0 +1,157 @@
...
+const mutations: MutationTree<MyModuleState> = {
+  setList (state, newList: string[]): void {
+    state.list = newList // state.list is typed using MyModuleState
+  }
+}
+```
+
+## Getters
+
+getters can be used with the `GetterTree<S, R>` interface which `S` is the current store type and `R` corresponds to the `RootState` and works exactly the same as a normal getter.
```
ederchrono:
```suggestion
Getters can be used with the `GetterTree<S, R>` interface which `S` is the current store type and `R` corresponds to the `RootState` and works exactly the same as a normal getter.
```

ktsn:
Thank you for contributing this docs! 
However, we are planing a new way to use Vuex with TypeScript and it will probably be different way with this. And I don't want to officially encourage this way to be honest as it is not type safe way and the most of people uses some wrapper lib to achieve it currently.

crutch12:
@ktsn Hi

Will we ever be able to use Vuex + Typescript without external libs like `vuex-class`?

For now even expression like `store.dispatch('space/action')` returns `any` (because first argument is string)

Do you plan to implement the new way to use of Vuex to support typing?

Thx

PR #1639, opened 2019-11-10
fix(types): avoid broadening vue instance type when using map helpers

### LINKED ISSUE #1119: Type checking is broken in Vue components when using mapping helpers

ktsn:
fix #1119

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "selenium-server": "^2.53.1",
     "terser": "^3.17.0",
     "todomvc-app-css": "^2.1.0",
-    "typescript": "^3.2.2",
+    "typescript": "^3.7.2",
     "vue": "^2.5.22",
     "vue-loader": "^15.2.1",
     "vue-template-compiler": "^2.5.22",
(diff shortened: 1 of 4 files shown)


DanielRuf:
I still get the issue with missing properties even if they are declared. Seems it does not use `this` from the component instance.

PR #1640, opened 2019-11-10
refactor: remove useless parameter on action handlers

### LINKED ISSUE #1578: where will we pass the second param cb for wrappedActionHandler?
### Version
3.1.1

### Reproduction link
[https://codepen.io/cucumbers/pen/bPQQeO](https://codepen.io/cucumbers/pen/bPQQeO)

### Steps to reproduce
in the function registerAction, we will wrap the true handler and push into entry.And the wrapped handler receive two params - payload and cb. But what we can see in function dispatch is that the second param cb will never be passed in.

### What is expected?
If the param cb  is useless, remove it.

The param cb for wrappedActionHandler is never used;

### LINKED ISSUE #1618: The third parameter of dispatch is undefined.
### Version
3.1.1

### Reproduction link
[https://codesandbox.io/s/vue-template-4fv3d](https://codesandbox.io/s/vue-template-4fv3d)

### Steps to reproduce
1. go to link;
2. open Console of page in the bottom;
3. see the undefined;

### What is expected?
In theory,it should show the code of cb.

In fact,it's undefined.

---
According to `Vuex` code, `wrappedActionHandler` function provide two params.One is `payload`,the other is `cb`.

```js
  // notice the cb
  entry.push(function wrappedActionHandler (payload, cb) {
    let res = handler.call(store, {
      dispatch: local.dispatch,
      commit: local.commit,
      getters: local.getters,
      state: local.state,
      rootGetters: store.getters,
      rootState: store.state
    }, payload, cb) // pass it and do nothing
    if (!isPromise(res)) {
      res = Promise.resolve(res)
    }
    if (store._devtoolHook) {
      return res.catch(err => {
        store._devtoolHook.emit('vuex:error', err)
        throw err
      })
    } else {
      return res
    }
  })

  // dispatch function only use payload
  dispatch (_type, _payload) {
    // check object-style dispatch
    const {
      type,
      payload
    } = unifyObjectStyle(_type, _payload)

    const action = { type, payload }
    const entry = this._actions[type]
    // ...
    const result = entry.length > 1
      ? Promise.all(entry.map(handler => handler(payload))) // only payload parameter
      : entry[0](payload)
  }
```
So I think `cb` parameter is useless and it should be delete from code in order to 
eliminate misunderstanding.

ktsn:
fix #1618 
fix #1578

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -433,7 +433,7 @@ function registerMutation (store, type, handler, local) {
 
 function registerAction (store, type, handler, local) {
   const entry = store._actions[type] || (store._actions[type] = [])
-  entry.push(function wrappedActionHandler (payload, cb) {
+  entry.push(function wrappedActionHandler (payload) {
     let res = handler.call(store, {
       dispatch: local.dispatch,
       commit: local.commit,
@@ -441,7 +441,7 @@ function registerAction (store, type, handler, local) {
       state: local.state,
       rootGetters: store.getters,
       rootState: store.state
-    }, payload, cb)
+    }, payload)
     if (!isPromise(res)) {
       res = Promise.resolve(res)
     }


michaeldjeffrey:
Thanks for removing this.

I had a memory when I started seeing this error after updating where to look. 

I was surprised to not see this in release notes though.

Issue #1647, opened 2019-11-18
PhantomJS is an unmaintained project


ossdev07:
### What problem does this feature solve?
The PhantomJS browser used by the vue package is an unmaintained project and also doesn't have support for aarch64.

I am curious to know that will you be interested in removing the phantomjs from vuex.
In my view, it will be better if the package uses all up to date dependencies.

Please revert that what are your thoughts on the phantomjs removal and share your suggestion on how to proceed further with this.

### What does the proposed API look like?
We could consider using the ChromeHeadless browser as an alternative to phantomjs as it is a maintained alternative to it.

ktsn:
Definitely we should do.

kiaking:
So I went ahead and created PR #1648 for this. Hope that will make sense.

ktsn:
fixed via #1648

PR #1648, opened 2019-11-19
#1647 Remove PhantomJS from E2E test, and use Chrom Headless Webdriver

kiaking:
Issue #1647

This PR removes PhantomJS from E2E test because it is not supported anymore. Instead, Chrome Headless Driver is used.

1. Remove PhantomJS from dependency and config.
2. Update Nightwatch and Chromedriver to the latest.
3. Remove Selenium too, since the latest Nightwatch can handle Chromesriver without Selenium, and also it's not recommended to use Selenium ([regarding the doc](https://nightwatchjs.org/gettingstarted/installation/#using-selenium-server)).
4. Tests adjusted a little since some part wasn't working correctly. Please refer to the inline comments on this.


test/e2e/specs/cart.js
```diff
@@ -20,7 +20,7 @@ module.exports = {
       .assert.containsText('.cart', 'H&M T-Shirt White - $10.99 x 1')
       .assert.containsText('.cart', 'Total: $1,011.01')
       .click('.cart button')
-      .waitFor(120)
+      .waitFor(200)
```
kiaking: I needed to change this because test was failing... Not sure why.

test/e2e/specs/todomvc.js
```diff
@@ -105,7 +105,7 @@ module.exports = {
       .dblClick('.todo:nth-child(1) label')
       .assert.count('.todo.editing', 1)
       .assert.focused('.todo:nth-child(1) .edit')
-      .clearValue('.todo:nth-child(1) .edit')
+    deleteValue('.todo:nth-child(1) .edit', 'test2')
```
kiaking: `clearValue` was not working. Not sure why but after `clearValue`, 2 todos was removed before `setValue`. So I'm guessing `clearValue` is emitting `blur` event or something...

So, I went ahead and created `deleteValue` function which will input "backspace" and deletes value. If this is the only workaround we have at the moment, I would like to extract that into a custom command so that we can do `browser.deleteValue`. Ref: [Custom Command Doc](https://nightwatchjs.org/guide/extending-nightwatch/#writing-custom-commands)

kiaking:
Updated Nightwatch to the latest version. Though it didn't fix `clearValue` problem 🤔

kiaking:
Hmmm... no idea why test is failing on circleci...

kazupon:
@kiaking 
vue-i18n `nightwatch.conf.js` might help you. 😎 
https://github.com/kazupon/vue-i18n/blob/dev/config/nightwatch.conf.js#L47-L56

kiaking:
@kazupon Ohhhh what the hell is that `headless` option lol  Thanks a lot! I'll try that and see how it goes!

kiaking:
OK I've optimized the Nightwatch config and circleci is padding, though I think it had nothing to do with it. Perhaps timing issue...

kiaking:
@ktsn Oh hi! What do you think about this?

PR #1653, opened 2019-12-01
refactor: remove extra parameter

zuochenxue:


ktsn:
Thanks for the PR. But it is actually referred in the function and I don't think accessing `store` in outside scope is a good idea.

Issue #1118, opened 2018-01-04
Clear all stores or set them all to its initial state


JonaMX:
### What problem does this feature solve?
I have several modules defined on my store, with its own store on each module:

```
export default new Vuex.Store({
  modules: {
    user,
    items,
    invoices,
    ... // A bunch of other modules
  }
})

```

The issue is that when the user logs out the app, all the information in the store remains since no mutation is being called to effect them or the set them to its initial state, therefore when another user logs in he can see information of the previous user, I know that I can just create a new mutation to set the store to its initial state, but that means that I'd need to call this mutation for each module that I have. 

Is there a way to clear them all?

### What does the proposed API look like?
It would be great to have a reserved mutation named 'clearAll' or alike, that resets all the module stores to its initial state as they were defined

```
...mapMutations([
  'clearAll'
])
```

Then on the logout method

```
logout: function () {
  this.$store.dispatch('logout').then(
    () => {
      this.$router.push('/login')
      this.clearAll()
    }
  )
},

```

robodude:
You can easily implement something like this on your own. 

On your 'root' Vuex store you can implement an action that calls each individual modules 'reset' mutation.

```
    actions: {
        clearAll({ commit }){
            commit("moduleA/reset")
            commit("moduleC/reset")
        }
    }
```

This gives you greater control over a generic solution because you might have modules that you wouldn't want to clear under some circumstances.

ktsn:
As @robodude described.

FYI, you can declare a function returning an initial state to reset a state.

```js
function initialState () {
  return { /* .. initial state ... */ }
}

export default {
  state: initialState,

  mutations: {
    reset (state) {
      // acquire initial state
      const s = initialState()
      Object.keys(s).forEach(key => {
        state[key] = s[key]
      })
    }
  }
}
```

DarynHolmes:
Thanks for this. I was wondering why does `reset` loop over all the keys? Why can't is just set `state` to `initialState()` ?

kiaking:
Because if you do that (`state = initialState`) then the state object will loose it's observer and it won't be reactive anymore. Same as plain object behavior in JavaScript.

```js
const object = { key: 'value' }

function one (o) {
  o.key = 'one'
}

one(object)

console.log(object) // <- { key: 'value' }

function two (o) {
  o = { key: 'two' }
}

two(object)

console.log(object) // <- { key: 'value' } Not modified!
```

peacefulseeker:
[...]

PR #1666, opened 2020-01-12
Store adds reset function that helps reset to initial state

huybuidac:
Currently, there is no true way to reset to initial state.

Adding Reset function to do that.

Fixes: https://github.com/vuejs/vuex/issues/1118

### DIFF
--- a/dist/vuex.common.js
+++ b/dist/vuex.common.js
@@ -1,6 +1,6 @@
 /**
  * vuex v3.1.2
- * (c) 2019 Evan You
+ * (c) 2020 Evan You
  * @license MIT
  */
 'use strict';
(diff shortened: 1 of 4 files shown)


ktsn:
We already can reset state without providing such functionality as I already mentioned in the original thread.
https://github.com/vuejs/vuex/issues/1118#issuecomment-356286218

huybuidac:
> We already can reset state without providing such functionality as I already mentioned in the original thread.
> [#1118 (comment)](https://github.com/vuejs/vuex/issues/1118#issuecomment-356286218)

It is a common part that every project uses and the implementation is really simple.
I don't know why Vuex makes all developers clone and clone the initState function and reset mutations.

therealcoder1337:
so we can already pass a function as state to reuse modules: https://vuex.vuejs.org/guide/modules.html#module-reuse
wouldn't it be easy to implement a `reset` functionality that builds upon this? this is really a major missing feature.

PR #1667, opened 2020-01-20
docs: Add clarity of module definition

Zhwt:
Add documentation clarity of module definition

### DIFF
--- a/docs/guide/modules.md
+++ b/docs/guide/modules.md
@@ -7,6 +7,7 @@ Due to using a single state tree, all state of our application is contained insi
 To help with that, Vuex allows us to divide our store into **modules**. Each module can contain its own state, mutations, actions, getters, and even nested modules - it's fractal all the way down:
 
 ``` js
+// both moduleA and moduleB are not instance of Vuex.Store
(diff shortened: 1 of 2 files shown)


kiaking:
I'm not sure about this comment since I think it's pretty obvious that modules are not a store instance. Do you find any particular place where it was misleading to let you think that modules ware store instance?

kiaking:
Will be closing this PR, as I mentioned above,  it's pretty obvious that modules are not a store instance. Thank you so much for the contribution though! 🤝

Zhwt:
@kiaking Thank you for your typing. Actually, this is not a big problem, and the documentation are not misleading at all 😂.
When I first using vuex, I made a mistake that I referenced modules as store instances, and I remember everything was fine until I called an action method, and the error messages were not related to incorrect type, took me about half an hour to debug and find out that I was using the wrong type.
Several days ago, some of my friends also run into this problem, so I decided to make a little change to the docs. One only need to take look at the documentation to correct the errors, but I think it would be better if we could notice them the differences before they make a mistake.

Issue #1591, opened 2019-07-30
Pass payload by value, not by reference


coreymcg:
### What problem does this feature solve?
Currently the payload passed to mutation functions is by reference, not value. This allows the mutation functions to directly update the payload data rather than the state object.

As an example see [editTodo() in the todo demo](https://github.com/vuejs/vuex/blob/91f3e69ed9e290cf91f8885c6d5ae2c97fa7ab81/examples/todomvc/store/mutations.js#L17):

```javascript
editTodo (state, { todo, text = todo.text, done = todo.done }) {
  todo.text = text
  todo.done = done
}
```

There are 2 downsides to this.

The first is that the function is mutating the payload and not the state. It is not clear here what is being mutated. We now need to traverse all around the app to find out.

If the payload was instead passed by value then updating the value would not actually do anything. All mutation functions would need to update the state object. This would make vuex stores more explicit and understandable as well as ensure there is a clear boundary between it and the app. 

The second is that Vuex plugins cannot log the mutations to a server and replay them back later. Here is a pseudocode example:

```javascript
// replaying back mutation logs
let logs = await getLogsFromServer();
for (log in logs) {
  // any editTodo mutations passed in below will not work
  // as the mutation will update the "log" object instead of 
  // the actual todo in the state object
  this.store.commit(log.type, log.payload);
}
```

I'm working on a plugin that does this (similar to [redux-vcr](https://github.com/joshwcomeau/redux-vcr)) and it would be great if there was a systematic guarantee that it would work with all Vuex stores. 

### What does the proposed API look like?
Example - No changes needed
```javascript
mutations: {
  // current: n is passed by reference
  increment (state, n) { 
    state.count  = n
  }
}

mutations: {
  // proposed: n is passed by value
  increment (state, n) { 
    state.count  = n
  }
}
```

Example B - editTodo

```javascript
mutations: {
  // current - can update payload
  editTodo (state, { todo, text = todo.text, done = todo.done }) {
    todo.text = text
    todo.done = done
  }
}

mutations: {
  // proposed - with pass by value we 
  // must update the state object
  editTodo (state, todo) {
     const index = state.todos.findIndex(t => t.id === todo.id)
     if (index > -1) state.todos.splice(index, 1, todo)
  }
}
```

kiaking:
Interesting thoughts! Well, we allow this usage because we sometimes want to directly mutate the returned value from a getter, which is not available in a mutation. I agree that this usage is not clear and a bit confusing.

[...]

PR #1670, opened 2020-01-27
#1591 refactor: do not mutate mutation payload in the todo app example

kiaking:
Issue #1591

This PR refactors the mutation logic in todo example to not mutate mutation payload directly in the simplest form.

It's still using the reference to todo (at `indexOf`) but I think it's more clearer than before that Vuex recommends to not mutate payload directly.

### DIFF
--- a/examples/todomvc/store/mutations.js
+++ b/examples/todomvc/store/mutations.js
@@ -15,7 +15,6 @@ export const mutations = {
   },
 
   editTodo (state, { todo, text = todo.text, done = todo.done }) {
-    todo.text = text
-    todo.done = done
+    state.todos.splice(state.todos.indexOf(todo), 1, { ...todo, text, done })
   }
 }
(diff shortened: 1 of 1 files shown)


kiaking:
Hmmm... test is failing. It passes at local env so it might be timing issue... We should look into the e2e test to be more stable.

kiaking:
OK now test is passing.

examples/todomvc/store/mutations.js
```diff
@@ -15,7 +15,6 @@ export const mutations = {
   },
 
   editTodo (state, { todo, text = todo.text, done = todo.done }) {
-    todo.text = text
-    todo.done = done
+    state.todos.splice(state.todos.indexOf(todo), 1, { ...todo, text, done })
```
ktsn: Can we refactor it for readability? Having everything on one line looks a bit too packed.

```suggestion
    const index = state.todos.indexOf(todo)
    state.todos.splice(index, 1, { 
      ...todo, 
      text, 
      done 
    })
```

kiaking: Ah of course 👍

kiaking: Done!

PR #1676, opened 2020-02-06
fix(#1571): remove undefined payload from dispatch

miguelrincon:
Fix for #1571

Removes unused payloads from the dispatch calls.

- Original PR at https://github.com/vuejs/vuex/pull/1573.
- Solution was proposed at https://github.com/vuejs/vuex/issues/1571#issuecomment-505838341

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -384,7 +384,7 @@ function makeLocalContext (store, namespace, path) {
         }
       }
 
-      return store.dispatch(type, payload)
+      return payload ? store.dispatch(type, payload) : store.dispatch(type)
     },
 
     commit: noNamespace ? store.commit : (_type, _payload, _options) => {
--- a/test/unit/store.spec.js
+++ b/test/unit/store.spec.js
@@ -77,6 +77,30 @@ describe('Store', () => {
     expect(store.state.a).toBe(3)
   })
 
+  it('dispatching actions without a payload', () => {
+    const store = new Vuex.Store({
+      state: {
+        a: false
+      },
+      mutations: {
+        [TEST] (state) {
+          state.a = true
+        }
+      },
+      actions: {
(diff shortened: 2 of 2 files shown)


kiaking:
@miguelrincon Thanks for the PR! Would it be possible to add test for this? (in case of regression

miguelrincon:
@kiaking Added a test but it's not what I expect.

I test for `arguments` in the spec, do you have a suggestion to fix? Is the spec wrong?

kiaking:
OK, maybe we have something wrong here. I've tested the following on current implementation.

```js
       state: {
         a: false
       },
       actions: {
         action ({ dispatch }) {
           console.log(arguments)
         }
       },
       modules: {
         b: {
           actions: {
             action ({ dispatch }) {
               console.log('action B', arguments)
             }
           }
         }
       }
     })
     store.dispatch('action')
     store.dispatch('b/action')     const store = new Vuex.Store({
       state: {
         a: false
       },
       actions: {
         action ({ dispatch }) {
           console.log(arguments)
         }
       },
       modules: {
         b: {
           actions: {
             action ({ dispatch }) {
               console.log('action B', arguments)
             }
           }
         }
       }
     })
     store.dispatch('action')
     store.dispatch('b/action')
```

Now, both `distapch('action')` and `dispatch('b/action')` got 2 arguments, one is the action context, and 2nd is `undefined` (payload). Which is not what #1571 was expecting. The former one should not receive the `undefined`, though it does.

The base issue's request was to make both static action and namespaced action to be consistent, and it is now.

Not how did this happened by maybe with some change we had since the issue was opened.

Therefore, I would say let's just skip this fix, and leave it as is, since it is now already having consistent behavior.

kiaking:
Closing due to the original issue has been closed.

PR #1678, opened 2020-02-09
refactor: remove extra loops in dispatch()

cjihrig:
This commit replaces two chained `filter()`-`forEach()` sequences in `dispatch()` with single `forEach()`s.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -136,8 +136,7 @@ export class Store {
     try {
       this._actionSubscribers
         .slice() // shallow copy to prevent iterator invalidation if subscriber synchronously calls unsubscribe
-        .filter(sub => sub.before)
-        .forEach(sub => sub.before(action, this.state))
+        .forEach(sub => { if (sub.before) sub.before(action, this.state) })
     } catch (e) {
       if (process.env.NODE_ENV !== 'production') {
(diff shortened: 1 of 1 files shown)


kiaking:
It looks good to me 😃  What do you think @ktsn ?

ktsn:
Well, I prefer the previous one as it is clear which part does what.

kiaking:
Hm, that true. Well yea if there're like so many subscriber attached, it might have some impact on performance, but I don't think there'll be such case (or at least it should be avoided) 🤔

I'm sorry to let you down, but what do you think @cjihrig ?

cjihrig:
I guess readability is subjective here. To me, personally, the version in this patch is more readable. It's also less code, one less loop over an array, and should consume less memory. It's up to you as the maintainers though.

kiaking:
I agree. The readability is subjective, indeed. I understand it's less loop, and it's a valid point. Though I guess this is the kind of thing that is pretty hard to make decisions. Like people could combine 2 functions into 1 and say it's less stack call.

As I mentioned, I think this 2 loop will not be a problem in performance nor memory usage. So, let's keep it as is.

It doesn't mean your suggestion is wrong or anything. I think we're all glad that you put your effort into making PR, and that's awesome. Thank you for your contribution!

PR #1687, opened 2020-03-02
docs: make obvious where to import store from

elevatebart:
As we browse through the modules page, it is unclear where the `store` is imported from.

Since it is a one liner, I thought it would clarify to have the import explicitly here.

Tell me what you think.

### DIFF
--- a/docs/guide/modules.md
+++ b/docs/guide/modules.md
@@ -280,7 +280,9 @@ export function createPlugin (options = {}) {
 You can register a module **after** the store has been created with the `store.registerModule` method:
 
 ``` js
-import { store } from 'vuex'
+import Vuex from 'vuex'
+
+const store = new Vuex.Store({ /* options */ })
 
 // register a module `myModule`
 store.registerModule('myModule', {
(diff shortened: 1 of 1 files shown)


ktsn:
You can't import `store` from `vuex` as it is an instance object which you create by `new Vuex.Store` and where users import from may be different between users.

elevatebart:
Thank you @ktsn. 
I did not know that, 
Do you think it is useful that I try to clarify it somewhere?

docs/guide/modules.md
```diff
@@ -280,6 +280,8 @@ export function createPlugin (options = {}) {
 You can register a module **after** the store has been created with the `store.registerModule` method:
 
 ``` js
+import { store } from 'vuex'
```
ktsn: Instead of importing, can we add a code to create a store module so that the user can know what is needed before using `registerModule`?

```suggestion
import Vuex from 'vuex'

const store = new Vuex.Store({ /* options */ })
```

elevatebart: Great idea

kiaking:
@elevatebart Hi! How is't going? Would it be possible for you to update the docs including comment from ktsn?

elevatebart:
御免なさい...

(^^;)

kiaking:
Thanks! Looking good to me. What do you say @ktsn ?

PR #1699, opened 2020-03-19
Throw an error when an unknown mutation is requested

andrewminer:
I noticed when I was running my unit tests that I got a number of `[vuex] unknown mutation type` errors in the log, but without any tests actually failing.  This was due to my having (deliberately) removed a store which provided those mutations.  This also removed those tests which were explicitly exercising that mutation, so there were no longer any tests which actually checked that the mutation worked.  So, that all checks out.

However...  there still were all the the places which were calling the now unknown mutation without triggering any errors of their own in _their_ tests.  This pull request changes the reaction to not finding a mutation to thow an error instead of swallowing the problem.  This will (correctly, to my mind) cause upstream code which erroneously commits a non-existant mutation to fail.

### DIFF
--- a/dist/vuex.common.js
+++ b/dist/vuex.common.js
@@ -385,7 +385,7 @@ Store.prototype.commit = function commit (_type, _payload, _options) {
   var entry = this._mutations[type];
   if (!entry) {
     if (process.env.NODE_ENV !== 'production') {
-      console.error(("[vuex] unknown mutation type: " + type));
+      throw new Error(("[vuex] unknown mutation type: " + type))
     }
     return
   }
--- a/src/store.js
+++ b/src/store.js
@@ -92,7 +92,7 @@ export class Store {
     const entry = this._mutations[type]
     if (!entry) {
       if (process.env.NODE_ENV !== 'production') {
-        console.error(`[vuex] unknown mutation type: ${type}`)
+        throw new Error(`[vuex] unknown mutation type: ${type}`)
       }
       return
     }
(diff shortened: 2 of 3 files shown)


kiaking:
Hi! Thank you so much for your contribution! It seems every Vue libs (including core) prefer to warn when asserting things.

So as for Vuex 3 & 4, we would like to keep things as is since this is technically a breaking change, and also, it aligns better with other Vue plibaries...

But I do see your point. We should definitely be more careful about this when designing Vuex 5.

andrewminer:
Thanks, @kiaking, I can certainly understand your point about this being a breaking change.  Perhaps there is some other way to allow people trying to test their code to capture this kind of error?  At the moment, the best one could do is to monkey-patch `console.log` and watch for error messages.  For example, here are a few potential ideas:

* Add a config value to the store which allows errors to be raised instead of log messages.
* Add a lifecycle callback which stores may choose to implement when this kind of thing happens.
* Add a config setting which causes a given mutation to be executed with the error as the payload.

I certainly don't want to break things for others, but it seems like there ought to be a non-monkey-patch way of ensuring unit tests fail when you've called a non-existant mutation even in Vue 3 / 4.
*

kiaking:
Adding a config option sounds interesting, though it's a kinda nasty workaround. Maybe this one could be made as a plugin that wraps `console.warn` and throws instead.

> I certainly don't want to break things for others, but it seems like there ought to be a non-monkey-patch way of ensuring unit tests fail when you've called a non-existant mutation even in Vue 3 / 4.

It's my personal opinion, but if your test calls non-existing mutations and if it doesn't fail, it sounds to me, the test is missing its assertion. Either checking the output (state change) or checking whether it's calling the correct mutation name should be asserted. Don't you think?

PR #1701, opened 2020-03-20
Fix getter type optional params

kootoopas:
Fix getter type to match implementation by marking optional params.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -103,7 +103,7 @@ export interface ActionObject<S, R> {
   handler: ActionHandler<S, R>;
 }
 
-export type Getter<S, R> = (state: S, getters: any, rootState: R, rootGetters: any) => any;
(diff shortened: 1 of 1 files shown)


kootoopas:
Node dependency error in deploy preview step:
```
5:34:36 PM: error mkdirp@1.0.3: The engine "node" is incompatible with this module. Expected version ">=10".
```

kiaking:
Thanks for the contribution! Though, all arguments to getters are not optional (`undefined`). The argument is always passed to the getters. You can try it out by console logging all arguments.

kootoopas:
I see, just checked it out...

kootoopas:
Although I still believe that the user should not be forced to provide all 4 args when they want to use a subset.

kiaking:
@kootoopas Maybe. But it's needed in fact. And the underscore prefix in TS such as `_getter` are there for that reason.

PR #1702, opened 2020-03-24
modify test description

sunnyke:
spell error:
file path:  /vuex/test/unit/util.spec.js 
diff: change nornal to normal


kiaking:
Thanks! Nice find 👍

PR #1708, opened 2020-04-07
refactor: normalizeMap uses Map instead of Object

baojie223:


kiaking:
@baojie223 Hi! Thanks for the PR, though why do you think we should be using Map instead of object in this context?

baojie223:
@kiaking Hi! Firstly, i think Map can bring us an unlimited key for future use. Secondly, it performs better in situations where key - value pairs are frequently added or deleted, but Object is not that good. So these are my reasons. : )

kiaking:
@baojie223 OK, sorry, I've checked and well, Vue supports IE9. And Vuex should also. Using Map will break that so we can't have this. Maybe in Vuex 4 we could, but we still have to wait until Vue 3 comes out with the final compatibilities.

Thank you so much for your contribution though! 🤝

PR #1711, opened 2020-04-10
refactor: Register nested modules through more uniform naming 

godky:
Usually people prefer to register nested modules `nested / myModule`  by passing in ` nested / myModule` instead of `['nested', 'myModule']`, which is more unified with the namespace.In addition,this modification will consider a single empty string path to be the root module.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -187,7 +187,7 @@ export class Store {
   }
 
   registerModule (path, rawModule, options = {}) {
-    if (typeof path === 'string') path = [path]
+    if (typeof path === 'string') path = path.split('/').filter(Boolean)
 
     if (process.env.NODE_ENV !== 'production') {
       assert(Array.isArray(path), `module path must be a string or an Array.`)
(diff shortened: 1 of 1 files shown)


kiaking:
@godky Thanks for the PR! This is quite interesting. I kinda like it, though not sure how big the benefit would be. Nice to have, but a bit more work on handling edge cases. At least this PR requires additional tests.

@ktsn What do you think of general ideas on this one?

ktsn:
Module namespace and path can be different, I would say we should keep current signature as it's path. Mixing them may be confusing.

godky:
I see。How about changing the parameter to `pathOrNamespace` and serialize it to path?

kiaking:
OK, maybe we should be closing this one for now as @ktsn mentioned. I think we could add features like this, but having many ways to define will cause another confusion as well. For example, people might be doing things like `registerModule(['foo', 'bar/hoo'])` and expect it to work. We could support such an edge case as well, though it just might not bring enough benefits.

I'll keep this one open for another week to see if we get any more feedbacks.

godky:
> OK, maybe we should be closing this one for now as @ktsn mentioned. I think we could add features like this, but having many ways to define will cause another confusion as well. For example, people might be doing things like `registerModule(['foo', 'bar/hoo'])` and expect it to work. We could support such an edge case as well, though it just might not bring enough benefits.
> 
> I'll keep this one open for another week to see if we get any more feedbacks.

I agree with you, thanks!

PR #1716, opened 2020-04-17
Use currentCount getter instead of accessing $store.state.count on Counter component

mohamedhayibor:


kiaking:
Thanks for the PR! Interesting adjustment.

While this might look better in the semantics point of view, maybe it makes more sense to keep it as is for this example.

By using `$store.state.count`, we show that it's OK to reference the store state inside component directly, and also explains how to do that. By abstracting it to be a getter, it might give the impression that users should always use getters to reference store state in components.

I think it's important to cover several use cases in examples like this one. What do you think?

mohamedhayibor:
I see. That's a very good point. I somehow got impression that accessing the store on components was a no go. My bad.

PR #1717, opened 2020-04-18
types: add typing support

### LINKED ISSUE #994: Feature: Manually install `vuex/types/vue.d.ts`

kiaking:
fix #994

Celebrating Vue Beta 2 release, now we can argument type our `$store` custom property in Vue Component 🎉  This PR adds typing support for Vuex 4.

- Add missing types, and update Vue interfaces.
- Upgrade TS to 3.8.
- Upgrade tsconfig to align with Vue 3 (support Map and such).
- Bring back type test in testing.


types/vue.d.ts
```diff
@@ -2,17 +2,17 @@
...
+declare module "@vue/runtime-core" {
+  interface ComponentCustomOptions {
     store?: Store<any>;
   }
 }
 
-declare module "vue/types/vue" {
-  interface Vue {
+declare module "@vue/runtime-core" {
+  interface ComponentCustomProperties {
```
ktsn, line 15: Do we remove store type from instance? We have an issue about this at #994

kiaking: Ah didn't see that issue, but yea, it makes sense. Let me remove the declaration from `ComponentCustomProperties`. We can keep `ComponentCustomOptions` since it doesn't matter so much right...? (it wouldn't be used by users to check store types anyway).

Do you think it can be applied to Vue 3 as well? Or would it be breaking change...? 🤔

types/index.d.ts
```diff
@@ -36,7 +38,7 @@ export declare class Store<S> {
   }): void;
 }
 
-export declare function install(Vue: typeof _Vue): void;
+export function createStore<S>(options: StoreOptions<S>): Store<S>;
```
chenxeed, line 41: Look forward to this!

Currently in the `alpha.1`, I'm facing this error on using `vuex` with typescript:

```
ERROR in /home/chenxeed/Projects/chenxeed/chenxeed-v2/src/store/index.ts(2,10):
2:10 Module '"../../node_modules/vuex/types"' has no exported member 'createStore'. Did you mean to use 'import createStore from "../../node_modules/vuex/types"' instead?
  > 1 | import { createStore } from "vuex";
```

I believe this should fix this error.

types/helpers.d.ts
```diff
@@ -1,12 +1,12 @@
...
+import { ComponentPublicInstance } from 'vue';
 import { Dispatch, Commit } from './index';
 
 type Computed = () => any;
 type InlineComputed<T extends Function> = T extends (...args: any[]) => infer R ? () => R : never
 type MutationMethod = (...args: any[]) => void;
 type ActionMethod = (...args: any[]) => Promise<any>;
 type InlineMethod<T extends (fn: any, ...args: any[]) => any> = T extends (fn: any, ...args: infer Args) => infer R ? (...args: Args) => R : never
-type CustomVue = Vue & Record<string, any>;
+type CustomVue = ComponentPublicInstance & Record<string, any>;
```
chenxeed, line 9: Shall this change also fix this error? :heart: 

```
Cannot use namespace 'Vue' as a type.
     7 | type ActionMethod = (...args: any[]) => Promise<any>;
     8 | type InlineMethod<T extends (fn: any, ...args: any[]) => any> = T extends (fn: any, ...args: infer Args) => infer R ? (...args: Args) => R : never
  >  9 | type CustomVue = Vue & Record<string, any>;

```

types/index.d.ts
kiaking: Yes, typings are not yet supported at `alpha.1`. This should do it 👍

types/helpers.d.ts
kiaking: YES!

types/vue.d.ts
ktsn: > We can keep ComponentCustomOptions since it doesn't matter so much right...?

Yes, I think so.

> Do you think it can be applied to Vue 3 as well?

Do you mean Vuex 3? As it's technically breaking change, we should only apply it for Vuex 4.

kiaking: Got ya. I'll also put this to my todo list to add in docs as well. I'll update the PR 👍

kiaking: Updated! I've removed the type test where checking `this.$store` as well.

kiaking:
I've added section in README that explains the new typings due to #994 change.

PR #1719, opened 2020-04-21
docs: unified the tips format

Jinjiang:
I found in docs, we both use `> note: xxx` and `:::tip xxx` so I unified them into `vuepress` specific format. Hope it would be better.
Thanks.


kiaking:
Nice 🙌  Yeah this make sense. Thanks a lot 👍

PR #1729, opened 2020-04-25
build: update the build system to the latest structure

kiaking:
Close #1253

This PR updates the build system to align with the latest Vue libs. Almost identical to how it is in Vuex 4. The differences are:

- Bundle name is as is.
- `UMD` build instead of `IIFE`.


kiaking:
Added global `__DEV__` support to prevent issue #1730 from happening.

.gitignore
```diff
@@ -1,12 +1,9 @@
...
-TODO.md
-lib
-docs/.vuepress/dist
-examples/**/build.js
-types/typings
-types/test/*.js
-explorations
+/docs/.vuepress/dist
+/examples/**/build.js
+/node_modules
```
ktsn, line 3:
```suggestion
node_modules
```

As `node_modules` should always be ignored.

kiaking: Fixed! I think this is same for Vuex 4 so I'll make sure to update that one as well when we merge this.

ktsn: Oh thanks. I had overlooked it 😣

rollup.config.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+  }
+
+  if (config.format === 'umd') {
+    c.output.name = c.output.name || 'Vuex'
+  }
+
+  c.plugins.push(replace({
+    __VERSION__: pkg.version,
+    __DEV__: config.format === 'es' && !config.browser
```
ktsn, line 40:
```suggestion
    __DEV__: config.format !== 'umd' && !config.browser
```

Commonjs build also need to have `process.env.NODE_ENV` in built file.

rollup.main.config.js
```diff
@@ -0,0 +1,10 @@
+import { createEntries } from './rollup.config'
+
+export default createEntries([
+  { input: 'src/index.js', file: 'dist/vuex.esm.browser.js', format: 'es', browser: true, transpile: false, env: 'development' },
+  { input: 'src/index.js', file: 'dist/vuex.esm.browser.min.js', format: 'es', browser: true, transpile: false, env: 'production' },
```
ktsn, line 5: Looks like we need `minify: true` here.

rollup.config.js
kiaking: Thanks! Fixed 👍

rollup.main.config.js
kiaking: Thanks! Fixed 👍

PR #1738, opened 2020-04-30
build: migrate tests to jest

kiaking:
Migrate test framework to Jest.

- All unit tests are written in Jest.
- Moved `setup` files out of `unit` filter and place it under `test`. This is for migrating E2E test to use Puppeteer instead of Nightwatch. In that time, we should be able to reuse this file since the test will all run with Jest.
- Updated Babel to the latest.
- Remove `babel-polyfill` due to deprecation. In the examples, we could add core-js pattern [described here at the top](https://babeljs.io/docs/en/babel-polyfill).
- Added `yarn coverage` to generate test coverage.


Kocal:
> This is for migrating E2E test to use prettier instead of Nightwatch

Do you mean Cypress and not Prettier?

kiaking:
> Do you mean Cypress instead of Prettier?

Sorry, I meant Puppeteer 🤦‍♂️. I love Cypress! Though we're trying align with Vue 3.

.babelrc
```diff
@@ -1,4 +1,4 @@
 {
-  "presets": ["env"],
-  "plugins": ["transform-object-rest-spread"]
+  "presets": ["@babel/preset-env"],
+  "plugins": ["@babel/plugin-proposal-object-rest-spread"]
```
ktsn, line 3: I think rest-spread plugin is already included in preset-env as it's in ES2018.

https://babeljs.io/docs/en/plugins#es2018

test/.eslintrc
```diff
@@ -0,0 +1,5 @@
+{
```
ktsn, line 1: Can we rename the file to `.eslintrc.json` as without extension is deprecated?
https://eslint.org/docs/user-guide/configuring#configuration-file-formats

.babelrc
kiaking: Thanks for pointing out. Yap, I've tested and indeed seems like we don't need it. Removed.

test/.eslintrc
kiaking: Uh I didn't know this one. Updated 👍

ktsn: Sorry we actually have two `.eslintrc`s. Can you update `test/.eslintrc` as well? 🙏

kiaking: Oops sorry my bad. Fixed that too 👍

kiaking:
@ktsn Updated and also rebased to the latest branch 👍

PR #1740, opened 2020-04-30
Allow action subscribers to catch rejections.

jmplahitko:
@kiaking - Thanks for the direction. This is the pull request to replace [#1531](https://github.com/vuejs/vuex/pull/1531)


cuebit:
Sorry @jmplahitko, I don't think we've been very clear in what we're asking of you, and I apologise for that.

When members ask to have `dist` removed from a commit, generally it means avoiding adding any files modified in the `dist` directory to the commit.

If it makes things easier, 1) take the `dist` directory from HEAD and throw it back in, 2) commit dist, and 3) run tests again but this time don't include any modifications in `dist`.

Hope that makes a little more sense.

jmplahitko:
@kiaking - Not your fault, I misunderstood. Hopefully that last commit gets us there. Thanks.

src/store.js
```diff
@@ -149,18 +149,32 @@ export class Store {
...
-      return res
+        resolve(res)
+      }, e => {
+        try {
+          this._actionSubscribers
+            .filter(sub => sub.error)
+            .forEach(sub => sub.error(action, this.state, e))
+        } catch (_e) {
+          if (__DEV__) {
+            console.warn(`[vuex] error in after action catch subscribers: `)
```
marswong, line 172: LGTM. To keep consistency make error message more semantic, use message `[vuex] error in error action subscribers: ` would be better :)

PR #1741, opened 2020-04-30
docs: module state declaration as function

cuebit:
This PR replaces all state declaration as functions for [module](https://vuex.vuejs.org/guide/modules.html) examples only.

Leading by example, pun intended, this will encourage users to avoid common pitfalls with shared declarative state.

Instance state examples are retained as `Object` since this does not apply to root state.

Closes #1477

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -185,7 +185,7 @@ const store = new Vuex.Store({ ...options })
   })
   ```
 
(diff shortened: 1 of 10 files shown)


cuebit:
@kiaking an afterthought, do the gitbook-docs require updating also?

kiaking:
@cuebit No. That's for old doc reference. Only updating `docs` folder is good 👍

cuebit:
@kiaking thought so. Thanks for confirming.

PR #1746, opened 2020-05-08
docs: add docs for the new `error` option for `subscribeAction` method

kiaking:
Add API docs for the new `error` option added to the `subscribeAction` method via #1740.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -233,7 +233,20 @@ const store = new Vuex.Store({ ...options })
   })
   ```
 
-  Most commonly used in plugins. [Details](../guide/plugins.md)
+  > New in 3.4.0
+
+  Since 3.4.0, `subscribeAction` can also specify the `error` option to catch the error thrown during the action dispatch. The `error` option will take an additional `error` object as the 3rd argument.
+
+  ``` js
+  store.subscribeAction({
+    after: (action, state, error) => {
+      console.log(`error action ${action.type}`)
+      console.error(error)
+    }
+  })
+  ```
+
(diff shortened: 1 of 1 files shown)


docs/api/README.md
```diff
@@ -233,7 +233,20 @@ const store = new Vuex.Store({ ...options })
...
+  > New in 3.4.0
+
+  Since 3.4.0, `subscribeAction` can also specify the `error` option to catch the error thrown during the action dispatch. The `error` option will take an additional `error` object as the 3rd argument.
+
+  ``` js
+  store.subscribeAction({
+    after: (action, state, error) => {
+      console.log(`error action ${action.type}`)
+      console.error(error)
+    }
```
cuebit, line 245: @kiaking the documentation doesn't seem to correlate with the code. This should be:

```js
error: (action, state, error) => {
  ...
}
```

docs/api/README.md
```diff
@@ -233,7 +233,20 @@ const store = new Vuex.Store({ ...options })
   })
   ```
 
-  Most commonly used in plugins. [Details](../guide/plugins.md)
+  > New in 3.4.0
+
+  Since 3.4.0, `subscribeAction` can also specify the `error` option to catch the error thrown during the action dispatch. The `error` option will take an additional `error` object as the 3rd argument.
```
cuebit, line 238: "3rd" should be "third" to keep consistent with other elongated argument descriptions.

docs/api/README.md
kiaking: Oops. Thanks a lot!

docs/api/README.md
kiaking: Thanks, fixed!

PR #1766, opened 2020-06-03
Add example for getters when using modules.

mprevdelta:
The notation becomes different for getters when a module is present and I believe an example would be very benefical.

### DIFF
--- a/docs/guide/getters.md
+++ b/docs/guide/getters.md
@@ -69,6 +69,14 @@ computed: {
 
 Note that getters accessed as properties are cached as part of Vue's reactivity system.
 
+### Property-Style Access with Modules
+
+Module properties will be merged into a single key of the form 'ModuleName/PropertyName'. Getters must be be accessed with bracket notation in this case:
+
+``` js
+store.getters['ModuleName/PropertyName']
+```
+
 ### Method-Style Access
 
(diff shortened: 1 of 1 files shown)


kiaking:
Thank you for the PR! While it might be nice to have this section, I think it's better to keep it as is since this is explained in "Modules" page. Unless you're using module, you don't need this syntax, and if we were to add this section to the getters page, we need to add to others (state, actions, mutations) as well.

mprevdelta:
I would in fact argue for updating those other pages with examples or cross references, since in learning Vue the module naming convention tripped me up in those instances also (I just didn't make a pull request).  Organizing the docs this way puts the onus on the reader to realize they have a special class of question--a "module question"--rather than a getter/setter/state/action/mutation question.

In the case of say, dispatching actions, I didn't find the situation so egregious because the method take a string and you know you're gonna have to pass some sort of special string:

```
store.dispatch('someAction')
store.dispatch('someModule.someAction') // nope
store.dispatch('someModule/someAction')  // aha!
```

...but in this case all of the examples lead the user to believe they'd use the dot notation.

`this.$store.getters.doneTodosCount`

At any rate, thank you for taking the time to review my request and for helping to make Vue a great tool.

kiaking:
I get your point. This is something people fall into quite often at the beginning. So having cross references, or link like "learn more about how to use getters with modules" might be nice to have 👍

PR #1779, opened 2020-06-17
Fix: Document return value when item was not found

localheinz:
This PR

* [x] documents the return value of `find` when an item was not found

💁‍♂️ I'm new to `vuejs` and `vuex`, so I'm not sure if this makes sense - but I could not find any documentation on what would be returned when an item is not found. What do you think?

### DIFF
--- a/test/unit/util.spec.js
+++ b/test/unit/util.spec.js
@@ -1,9 +1,14 @@
 import { find, deepCopy, forEachValue, isObject, isPromise, assert } from '@/util'
 
 describe('util', () => {
-  it('find', () => {
+  it('find: returns item when it was found', () => {
     const list = [33, 22, 112, 222, 43]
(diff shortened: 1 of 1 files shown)


cuebit:
`find` is an internal utility helper so there’s no strict assertion on the input and output type for this function. I’m not entirely sure what this PR addresses for user land.

That being said, analysing the code, it’s pretty straightforward to determine that a non-existent index will always return undefined. That’s something beyond vuex and is fundamentally part of JavaScript.

kiaking:
Yeah find returns `undefined` like `Array.prototype.find` does. Though, I guess having good description on test case is good, and it might improve the test coverage so why not. Thanks for the PR 🙌

localheinz:
Thank you, @cuebit and @kiaking!

PR #1782, opened 2020-06-19
docs(guide/modules): Add examples for mapGetters with namespace

michapietsch:
Hey,

I found myself looking up the exact syntax repeatedly, so I'd like to contribute these examples.

Best regards,
Micha


kiaking:
Uh thanks a lot! yeah this does make sense 👍

PR #1786, opened 2020-06-29
fix: warn when unregistering non existing module

kiaking:
close #1426

This PR adds warning when removing non existing module. Currently, it fails due to `runtime` not exist in `undefined` error. It's follow up PR for #1426.

It will only warn in development mode, so it will not introduce any braking changes.

### DIFF
--- a/src/module/module-collection.js
+++ b/src/module/module-collection.js
@@ -49,8 +49,9 @@ export default class ModuleCollection {
   unregister (path) {
     const parent = this.get(path.slice(0, -1))
     const key = path[path.length - 1]
+    const child = parent.getChild(key)
 
-    if (!parent.getChild(key)) {
+    if (!child) {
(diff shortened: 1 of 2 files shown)


src/module/module-collection.js
```diff
@@ -49,7 +49,20 @@ export default class ModuleCollection {
   unregister (path) {
     const parent = this.get(path.slice(0, -1))
     const key = path[path.length - 1]
-    if (!parent.getChild(key).runtime) return
+
+    if (!parent.getChild(key)) {
```
ktsn, line 53: Can we store the return value of `getChild` in a variable so that we reuse it with the next if condition?

kiaking: Oh yeah good point! Thanks. I've done it, and rebased (to the latest dev) and pushed 👍

PR #1793, opened 2020-07-05
fix code inconsistencies

samuelgozi:
I fixed some code inconsistencies in the English documentation.
Please let me know if ane changes are needed.

### DIFF
--- a/docs/guide/README.md
+++ b/docs/guide/README.md
@@ -27,7 +27,7 @@ const store = new Vuex.Store({
     count: 0
   },
   mutations: {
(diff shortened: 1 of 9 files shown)


cuebit:
Thanks for the PR, @samuelgozi. Unfortunately this PR does not qualify as fixing anything, rather opinionated formatting of the code examples. To clarify, the use of ES6 syntax and the chosen code format in the docs need to be consistent with that of the codebase. We do not have any intention of introducing additional formatting rules in the active branch at this time.

PR #1797, opened 2020-07-12
Support modules and state functions in Store typings

zOadT:
_DISCLAIMER: The use of [conditional types](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#conditional-types) and [`ReturnType<T>`](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-8.html#predefined-conditional-types) in this PR requires the user to use at least Typescript 2.8. Right now the use of [generic parameter defaults](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults) in helper.d.ts only requires Typescript 2.3._

This PR improves the typings of `Store`. Right now, given code like

```js
let moduleC = {
    state() {
        return {
            valueC: -13
        }
    }
}
let moduleA = {
    state() {
        return ({
            valueA: "6",
        })
    },
    modules: {
        c: moduleC,
    }
}
let moduleB = {
    state: {
        a: 36,
    }
}
let store = new Vuex.Store({
    state: () => ({
        foo: 6
    }),
    modules: {
        a: moduleA,
        b: moduleB,
    }
})
```

would result in the type of `store.state` to only consider the local state in the root. Also `store.state` would be typed as a function and not as its return type. With this PR the type of `store.state` will correspond to

```js
{
    foo: number;
    a: {
        valueA: string;
        c: {
            valueC: number;
        }
    };
    b: {
        a: number;
    };
}
```

### Compatibility
We provided defaults for new generics, therefore one should be able to still use the types as before.

### About older TS versions
There seems to be a strange bug in older versions of Typescript. Sometimes the type of a store differs depending on wheather you use the `{ state() { return {...}}, ... }` or the `{ state: () => ({...}), ... }` syntax. Its not that much of a problem though. We tested the typings in 2.8.4 and the only thing that happens was that the properties are missing, but TS does not throw an error (as long as you don't try to access one of these properties) In newer versions of Typescript everything is fine.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -15,17 +15,17 @@ export declare class Store<S, T extends ModuleTree<S> = {}> {
   readonly state: State<S, T>;
   readonly getters: any;
 
-  replaceState(state: S): void;
+  replaceState(state: State<S, T>): void;
 
   dispatch: Dispatch;
   commit: Commit;
 
-  subscribe<P extends MutationPayload>(fn: (mutation: P, state: S) => any, options?: SubscribeOptions): () => void;
-  subscribeAction<P extends ActionPayload>(fn: SubscribeActionOptions<P, S>, options?: SubscribeOptions): () => void;
-  watch<T>(getter: (state: S, getters: any) => T, cb: (value: T, oldValue: T) => void, options?: WatchOptions): () => void;
+  subscribe<P extends MutationPayload>(fn: (mutation: P, state: State<S, T>) => any, options?: SubscribeOptions): () => void;
+  subscribeAction<P extends ActionPayload>(fn: SubscribeActionOptions<P, State<S, T>>, options?: SubscribeOptions): () => void;
+  watch<U>(getter: (state: State<S, T>, getters: any) => U, cb: (value: U, oldValue: U) => void, options?: WatchOptions): () => void;
 
-  registerModule<T>(path: string, module: Module<T, S>, options?: ModuleOptions): void;
-  registerModule<T>(path: string[], module: Module<T, S>, options?: ModuleOptions): void;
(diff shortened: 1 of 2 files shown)


kiaking:
Thank you so much for the PR! Yeah this kinda makes sense. It doesn't address the issue of dynamic module registration but seems like it's better than the current where we have to specify whole module tree interface manually.

@ktsn  what do you think about this? If we don't find any risk, maybe it's nice to have?

zOadT:
BTW sorry for messing up the commit names. I can create a new PR with proper commit names and tests if desired.

Concerning dynamic module registration: I don't think it is possible to do it all automatically there, except maybe `store.registerModule` would return the store instance (this way we we could return the store with updated typings).

kiaking:
Commit massage should be fine, we can squash and rename it when merging the PR 👍 Though the test code should be added before merging.

But, please let's wait until we get response from Katashin, I would like to make sure this is the right direction we wanna take 🙇

pulkit-github (approved):

kiaking:
@ktsn Hi! Just checking in here 😳 Any thought on this issue?

ktsn:
It looks fine to me.

zOadT:
Thanks for your follow up! 🙂

I will apply the state typing to method parameters (where possible) and will update the tests ASAP

zOadT:
The typings are used by more methods and tests for the new typings are added.

I think this PR is ready to merge, but please feel free to tell me if there is anything I should alter 🙂

Greendogo:
When do proposals like this get evaluated for merging?  The functionality described here seems pretty important.

PR #1801, opened 2020-07-24
Update getters.js

7006:
rewrite `recentHistory` getter using negative index for`start` argument of Array.prototype.slice 

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/slice


kiaking:
Thanks for the PR! Nice, yeah this looks much cleaner 👍

PR #1803, opened 2020-07-27
fix(types): add lost argument of useStore

### LINKED ISSUE #1807: useStore() type definition is missing parameter for storeKey.
### Version
4.0.0-beta.4

### Steps to reproduce
In the Vue4 type definition the useStore function is defined as

https://github.com/vuejs/vuex/blob/83bc26fe55417692f386db9b76634eb58794282e/types/index.d.ts#L48

but in the source you can provide a key, to enable more then one store in the application.

https://github.com/vuejs/vuex/blob/83bc26fe55417692f386db9b76634eb58794282e/src/injectKey.js#L5

### What is expected?
The type definition should match the function

It does not match ;) so only one store can be used with typescript

euaaaio:
This PR adds lost argument `key` to `useStore` function types.
Resolve #1807

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -45,7 +45,7 @@ export declare class Store<S> {
 
 export function createStore<S>(options: StoreOptions<S>): Store<S>;
 
-export function useStore<S = any>(): Store<S>;
+export function useStore<S = any>(key?: string): Store<S>;
 
 export interface Dispatch {
   (type: string, payload?: any, options?: DispatchOptions): Promise<any>;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -45,7 +45,7 @@ export declare class Store<S> {
 
 export function createStore<S>(options: StoreOptions<S>): Store<S>;
 
-export function useStore<S = any>(injectKey?: InjectionKey<Store<any>> | string): Store<S>;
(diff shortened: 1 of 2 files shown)


kiaking:
Thanks for the PR! Yap, this is a defect. However, could you add support for `InjectionKey` as well as String?
Ref: https://composition-api.vuejs.org/api.html#dependency-injection

So It should be I think?

```js
export function useStore<S = any>(key?: InjectionKey<Store<S>> | string): Store<S>;
```

euaaaio:
@kiaking, you're right. I refactored and added a few tests.

types/index.d.ts
```diff
@@ -45,7 +45,7 @@ export declare class Store<S> {
 
 export function createStore<S>(options: StoreOptions<S>): Store<S>;
 
-export function useStore<S = any>(): Store<S>;
+export function useStore<S = any>(injectKey?: InjectionKey<Store<any>> | string): Store<S>;
```
kiaking, line 48: I think it should be ` InjectionKey<Store<S>>`?

```js
export function useStore<S = any>(injectKey?: InjectionKey<Store<S>> | string): Store<S>;
```

euaaaio: 😳

PR #1814, opened 2020-08-12
added code of conduct to project

jivthesh:


kiaking:
Hi Thanks for the PR! Currently we already linked the Code of Conduct to the Vue's repo from https://github.com/vuejs/vuex/blob/dev/.github/contributing.md. This is what we do in Vuex Next and Due Router too so I think we should keep it this way.

Thank you for taking your time to do this! 🙌

PR #1817, opened 2020-08-16
docs(typo): fix punctuation

jamiedishy:
The period should be inside the quotations, not outside the quotations.


cuebit:
Thanks so much 🙌

PR #1819, opened 2020-08-20
Update shop.js

Amikawayon:
cb products arg


arunkp (approved):
:+1:

kiaking:
Hi Thanks for the PR! Could you elaborate a bit more on this PR? I don't think we're using that `product` argument anywhere anyway...? Why do we need to pass `products` to the callback?

kiaking:
Closing due to inactivity. Please feel free to reopen this if there's anything 👍

PR #1834, opened 2020-09-19
docs: show mutations in 4.x example also

abhineet97:
In the example for Vuex 3.x, mutations have been illustrated. I've added them to the example for 4.x too for consistency.

### DIFF
--- a/docs/guide/README.md
+++ b/docs/guide/README.md
@@ -47,6 +47,11 @@ const store = createStore({
     return {
       count: 1
     }
+  },
+  mutations: {
+    increment (state) {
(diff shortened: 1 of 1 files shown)


cuebit:
I think the same argument could be made for getters and actions, and not mutations alone. 

However, at this time, the README exhibits procedural changes to get started with Vuex 4 such as instantiating a store and mounting it in Vue 3. It's not intended to outline features in their entirety since the API is unchanged in both Vuex 3 and 4. Thought, this will be covered in the official docs for Vuex 4. 👍

PR #1838, opened 2020-09-30
docs: warn about calling unsubscribe (fix #1050)

### LINKED ISSUE #1050: Add documentation entry for watch store unsubscription on component
Actually there are no information regarding the necessity to "unsubscribe" store watch on component destroy.

It would be good to set a warning on the API documentation (https://vuex.vuejs.org/en/api.html) to let users know that without explicit unsubscription, the watch will be keep even after component destroy.

This issue has been reported here https://github.com/vuejs/vuex/issues/599 and closed as a won't fix.

This problem cause *memory leak* and in most of the case *errors* extremely hard to debug.

## For a minimalist resolution
A *warning quote* in https://vuex.vuejs.org/en/api.html .

## Nice to have
A new entry in menu to explain how to use `watch` (and then the tips for the *unsubscription*).
It could be great to give a better knowledge of this feature.

anschwa:
I added a warning text to the docs about remembering to call `unsubscribe` when subscribing to the store.

### DIFF
--- a/docs/api/README.md
+++ b/docs/api/README.md
@@ -176,13 +176,20 @@ const store = new Vuex.Store({ ...options })
 
 -  `subscribe(handler: Function, options?: Object): Function`
 
-  Subscribe to store mutations. The `handler` is called after every mutation and receives the mutation descriptor and post-mutation state as arguments:
+  Subscribe to store mutations. The `handler` is called after every mutation and receives the mutation descriptor and post-mutation state as arguments.
+
+:::warning WARNING
+A subscription created within a Vue component will remain after the component destroyed. You must call `unsubscribe` to prevent resource leaks.
+:::
 
   ``` js
-  store.subscribe((mutation, state) => {
+  const unsubscribe = store.subscribe((mutation, state) => {
     console.log(mutation.type)
     console.log(mutation.payload)
   })
+
+  // remember to call unsubscribe
+  unsubscribe()
   ```
 
   By default, new handler is added to the end of the chain, so it will be executed after other handlers that were added before. This can be overridden by adding `prepend: true` to `options`, which will add the handler to the beginning of the chain.
@@ -201,13 +208,20 @@ const store = new Vuex.Store({ ...options })
 
   > New in 2.5.0
 
-  Subscribe to store actions. The `handler` is called for every dispatched action and receives the action descriptor and current store state as arguments:
+  Subscribe to store actions. The `handler` is called for every dispatched action and receives the action descriptor and current store state as arguments.
+
+:::warning WARNING
+A subscription created within a Vue component will remain after the component destroyed. You must call `unsubscribe` to prevent resource leaks.
+:::
 
   ``` js
-  store.subscribeAction((action, state) => {
+  const unsubscribe = store.subscribeAction((action, state) => {
     console.log(action.type)
     console.log(action.payload)
   })
+
+  // remember to call unsubscribe
+  unsubscribe()
   ```
 
   By default, new handler is added to the end of the chain, so it will be executed after other handlers that were added before. This can be overridden by adding `prepend: true` to `options`, which will add the handler to the beginning of the chain.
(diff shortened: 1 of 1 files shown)


docs/api/README.md
```diff
@@ -176,13 +176,20 @@ const store = new Vuex.Store({ ...options })
 
 -  `subscribe(handler: Function, options?: Object): Function`
 
-  Subscribe to store mutations. The `handler` is called after every mutation and receives the mutation descriptor and post-mutation state as arguments:
+  Subscribe to store mutations. The `handler` is called after every mutation and receives the mutation descriptor and post-mutation state as arguments.
+
+:::warning WARNING
```
kiaking, line 181: I think when you actually want to unsubscribe when unregistering the Vuex modules. So instead of making this "warning", how about we just add a description to the `unsubscribe` function?

---

Subscribe to store mutations. The `handler` is called after every mutation and receives the mutation descriptor and post-mutation state as arguments.

The `subscribe` method will return `unsubscribe` function, which can be used to stop the subscription.

kiaking (changes requested):
Hi! Thanks for the PR! I made a comment, could you take a look?

docs/api/README.md
anschwa: Thanks! I understand that Vuex is quite decoupled from Vue components, but I think mentioning the need to call unsubscribe when called from a component is still helpful. 

When I was debugging a leak caused by one of our components calling `this.$store.subscribeAction` on create, it wasn't obvious the subscription wasn't cleaned up after being destroyed. I pushed up a change that removes the warning block and extends your suggestion with:
> The `subscribe` method will return an `unsubscribe` function, which should be called when the subscription is no longer needed. For example, when unregistering a Vuex module or before destroying a Vue component.

I will make any additional changes as needed.

docs/api/README.md
```diff
@@ -176,13 +176,17 @@ const store = new Vuex.Store({ ...options })
 
 -  `subscribe(handler: Function, options?: Object): Function`
 
-  Subscribe to store mutations. The `handler` is called after every mutation and receives the mutation descriptor and post-mutation state as arguments:
+  Subscribe to store mutations. The `handler` is called after every mutation and receives the mutation descriptor and post-mutation state as arguments.
+  The `subscribe` method will return an `unsubscribe` function, which should be called when the subscription is no longer needed. For example, when unregistering a Vuex module or before destroying a Vue component.
```
kiaking, line 180: I think we should be a little more explicit about the use cases here. Because people might get confused about Vuex module, or Vue Component mentioned here since there is no example nor description provided formerly about such use cases. So how about we make it like this.

First of all, let's move `unsubscribe` feature description at the last of the section (replace "To stop subscribing, call the returned unsubscribe function." text). Then add:

---

The `subscribe` method will return an `unsubscribe` function, which should be called when the subscription is no longer needed. For example, you might subscribe to a Vuex Module, and then later, you want to unregister the module, or you might call `subscribe` method inside Vue Component and then destroy the component later. In these cases, you should remember to unsubscribe the subscription manually.

docs/api/README.md
```diff
@@ -176,13 +176,17 @@ const store = new Vuex.Store({ ...options })
...
 
   ``` js
-  store.subscribe((mutation, state) => {
+  const unsubscribe = store.subscribe((mutation, state) => {
     console.log(mutation.type)
     console.log(mutation.payload)
   })
+
+  // remember to call unsubscribe
```
kiaking, line 188: I think this should be a more informative note. Because it's not always necessary to unsubscribe. How about we make this:

you may call `unsubscribe` method later to stop the subscription

kiaking (changes requested):
Thanks for the clarification! I guess it might be confusing due to that Vue Component's `computed`, or `watch` option doesn't require manual unsubscribe.

I like the direction of the docs. I've made a few more comments. Please have a look 🙏

docs/api/README.md
kiaking: Made new comment! Please have a look 🙏

anschwa:
@kiaking let me know if there any additional changes you need.

kiaking (approved):
Thanks a lot! Looking good I think 👍  Appreciate so much for your work 🤝

PR #1853, opened 2020-10-18
feat(types): adding logger type for logger plugin

itomtom:
* createLogger options takes in logger which by default is console but
can be overridden with own logger.
* Adding Ilogger for logger options params which expects log method to
be implemented

### DIFF
--- a/types/logger.d.ts
+++ b/types/logger.d.ts
@@ -1,5 +1,10 @@
 import { Payload, Plugin } from "./index";
 
+interface Logger extends Partial<Pick<Console, 'groupCollapsed' | 'group' | 'groupEnd'>> {
+  log(message: string, colour: string, action: any): void;
+  log(message: string): void;
+}
+
 export interface LoggerOption<S> {
   collapsed?: boolean;
   filter?: <P extends Payload>(mutation: P, stateBefore: S, stateAfter: S) => boolean;
@@ -9,6 +14,7 @@ export interface LoggerOption<S> {
   actionTransformer?: <P extends Payload>(action: P) => any;
   logMutations?: boolean;
   logActions?: boolean;
+  logger?: Logger;
 }
 
 export default function createLogger<S>(option?: LoggerOption<S>): Plugin<S>;
--- a/types/test/index.ts
+++ b/types/test/index.ts
@@ -437,10 +437,17 @@ namespace Plugins {
     });
   }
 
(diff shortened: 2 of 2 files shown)


types/logger.d.ts
```diff
@@ -1,5 +1,10 @@
 import { Payload, Plugin } from "./index";
 
+interface ILogger extends Partial<Console> {
```
kiaking, line 3: Can we name this just `Logger` instead `I` prefix?

Also, would it be possible to use `Pick` instead of `Partial`, since we require few more methods for the interface.

```ts
interface Logger extends Pick<Console, 'log' | 'group' | 'groupCollapsed' | 'groupEnd'> {}
```

kiaking (changes requested):
Thank you so much for the PR! yeah we totally forgot about `logger` option hah...

I've made a comment. Could you have a look :pray: ?

types/logger.d.ts
itomtom: I had to make it `Partial<Pick<Console, 'log' | 'group' | 'groupCollapsed' | 'groupEnd>>` since **group**, **groupCollapsed** and **groupEnd** is not compulsory

kiaking: Thanks! OK, yea I think what you mean by those functions are not compulsory, because it would not cause any failure even those methods are not implemented. Hmmm. Not sure what is the best practice in these cases, but I think it make sense with your implementation.

However, I think we should update the interface for the `log` method to match `Console.log` interface. What do you think? And in this case, I think we can strip `log` from `Pick`.

```ts
interface Logger extends Partial<Pick<Console, 'groupCollapsed' | 'group' | 'groupEnd'>> {
  log(...data: any[]): void;
}
```

itomtom: Instead of `log(...data: any[]): void` I went with 
```typescript
interface Logger extends Partial<Pick<Console, 'groupCollapsed' | 'group' | 'groupEnd'>> {
  log(message: string, colour: string, action: any): void;
  log(message: string): void;
}
```
to be a bit more explicit

itomtom: I just remember if the world of JavaScript the signatures are too important so `log(...data: any[]): void` should work as well

kiaking: Ah cool thanks!

kiaking (approved):

PR #1872, opened 2020-11-13
Add Node.js esm entry point

sebamarynissen:
This PR adds a Node.js esm entry point for Vuex. Currently it is already possible to do
```js
import Vuex from 'vuex'
```
on Node 13.2 and up, but unfortunately it's not possible to do
```js
import { mapGetters } from 'vuex';
```
because Node doesn't support named imports from CommonJS (note: it actually *does*, but only if the imported cjs file adheres to some rules, such as using `exports.mapGetters`). This PR adds support for named imports on Node as well.

While it *is* possible to directly use `dist/vuex.esm.js`, this PR does not use that file but chooses to use an esm wrapper file as to avoid the [dual package hazard](https://nodejs.org/api/packages.html#packages_approach_1_use_an_es_module_wrapper).

### Why?

Node.js now fully supports esm so let's encourage the migration process by improving interoperability of the code we write. I've created this PR in the wake of writing [vue-esm-loader](https://www.npmjs.com/package/vue-esm-loader) which allows importing `.vue` single-file components on Node without a build step. I noticed that I was unable to test my Vue components because they were using
```js
import { mapState } from 'vuex';

export default {
  computed: {
    ...mapState(['state']),
  },
};
```
and hence decided to add this functionality to Vuex.

Related: https://github.com/chaijs/chai/pull/1340


package.json
```diff
@@ -3,6 +3,13 @@
   "version": "3.5.1",
   "description": "state management for Vue.js",
   "main": "dist/vuex.common.js",
+  "exports": {
+    ".": {
+      "require": "./dist/vuex.common.js",
+      "import": "./src/index.mjs"
```
kiaking, line 9: I think we should place `index.mjs` at `dist`, since `src` is not packed as NPM package. We should be able to add copying script at `scripts/build.js` to simply copy `index.mjs` from `src` dir.

kiaking (changes requested):
Thanks a lot on the PR! I've tested this at my environment too and it's working. Nice one 👍 

I left a comment on not referencing `src` folder. Please have a look 🙏

package.json
sebamarynissen: Oops, didn't take that one into account. I pushed a new commit that copies `index.mjs` into the `dist/` folder while also renaming it to `vuex.mjs` to be better in line with the other files in the folder. I have to mention that I still kept the import path like this:
```js
import Vuex from '../dist/vuex.common.js'
```
Strictly spoken this can now be reduced to
```js
import Vuex from './vuex.common.js'
```
but I chose not to do it so that `import { mapGetters } from 'vuex/index.mjs'` still works in development as well.

kiaking: Very nice! That's perfect 👍

kiaking (approved):

PR #1883, opened 2020-11-22
fix: fix getters being destroyed on component destroy (#1878)

### LINKED ISSUE #1878: lost previous registered module during multiple registration
### Version
4.0.0-rc.1

### Reproduction link
[https://github.com/musashiM82/register-module-in-before-create-page](https://github.com/musashiM82/register-module-in-before-create-page)

### Steps to reproduce
* clone repo
* execute `npm i && npm run serve`
* open web app in browser
* in "Home" page you can see "state.home.count 1"
* click on "About" link
* click on "Home" link
* in "Home" page you can see "state.home.count" now

### What is expected?
it is expected that the second time that I go into Home page I continue to see "state.home.count 1"

it is happening that I see "state.home.count". it is like the previous registered module is lost

kiaking:
fix #1878

This PR aims to fix #1878, but at the moment this is just adding failing test case.

The problem we have is that when component gets destroyed, the getters (`computed`) get's destroyed via Vue. Hence if we call `registerModule` inside a component, any newly registered `computed` will disappear when that component gets destroyed, for example, when switching route.

The issue: vuejs/vue-next#1532
Solution RFC: vuejs/rfcs#212

We have to wait for Vue ~3.1~ 3.2 to completely resolve this issue.

### DIFF
--- a/src/store-util.js
+++ b/src/store-util.js
@@ -1,4 +1,4 @@
-import { reactive, watch } from 'vue'
+import { reactive, computed, watch, effectScope } from 'vue'
 import { forEachValue, isObject, isPromise, assert, partial } from './util'
 
 export function genericSubscribe (fn, subs, options) {
@@ -29,6 +29,7 @@ export function resetStore (store, hot) {
 
 export function resetStoreState (store, state, hot) {
   const oldState = store._state
+  const oldScope = store._scope
 
   // bind store public getters
   store.getters = {}
@@ -36,22 +37,33 @@ export function resetStoreState (store, state, hot) {
   store._makeLocalGettersCache = Object.create(null)
   const wrappedGetters = store._wrappedGetters
   const computedObj = {}
-  forEachValue(wrappedGetters, (fn, key) => {
-    // use computed to leverage its lazy-caching mechanism
-    // direct inline function use will lead to closure preserving oldState.
-    // using partial to return function with only arguments preserved in closure environment.
-    computedObj[key] = partial(fn, store)
-    Object.defineProperty(store.getters, key, {
-      // TODO: use `computed` when it's possible. at the moment we can't due to
-      // https://github.com/vuejs/vuex/pull/1883
-      get: () => computedObj[key](),
-      enumerable: true // for local getters
+  const computedCache = {}
+
+  // create a new effect scope and create computed object inside it to avoid
+  // getters (computed) getting destroyed on component unmount.
+  const scope = effectScope(true)
+
+  scope.run(() => {
+    forEachValue(wrappedGetters, (fn, key) => {
+      // use computed to leverage its lazy-caching mechanism
+      // direct inline function use will lead to closure preserving oldState.
+      // using partial to return function with only arguments preserved in closure environment.
+      computedObj[key] = partial(fn, store)
+      computedCache[key] = computed(() => computedObj[key]())
+      Object.defineProperty(store.getters, key, {
+        get: () => computedCache[key].value,
+        enumerable: true // for local getters
+      })
     })
   })
 
   store._state = reactive({
     data: state
   })
+
+  // register the newly created effect scope to the store so that we can
+  // dispose the effects when this method runs again in the future.
+  store._scope = scope
 
   // enable strict mode for new state
   if (store.strict) {
@@ -66,6 +78,11 @@ export function resetStoreState (store, state, hot) {
         oldState.data = null
       })
     }
+  }
+
+  // dispose previously registered effect scope if there is one.
+  if (oldScope) {
+    oldScope.stop()
   }
 }
 
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -4,7 +4,7 @@ import puppeteer from 'puppeteer'
 export function mount (store, component) {
   const el = createElement()
 
-  component.render = () => {}
+  component.render = component.render ? component.render : () => {}
 
   const app = createApp(component)
 
--- a/test/helpers.js
+++ b/test/helpers.js
@@ -4,7 +4,7 @@ import puppeteer from 'puppeteer'
 export function mount (store, component) {
   const el = createElement()
 
-  component.render = component.render ? component.render : () => {}
+  component.render = component.render || (() => {})
 
   const app = createApp(component)
 
--- a/test/unit/modules.spec.js
+++ b/test/unit/modules.spec.js
@@ -126,7 +126,7 @@ describe('Modules', () => {
       expect(mutationSpy).toHaveBeenCalled()
     })
 
-    it.only('should keep getters when component gets destroyed', async () => {
+    it('should keep getters when component gets destroyed', async () => {
       const store = new Vuex.Store()
 
       const moduleA = {
@@ -165,7 +165,7 @@ describe('Modules', () => {
       await nextTick()
 
       store.commit('moduleA/increment')
-      console.log(store.state.moduleA.value)
+
       expect(store.getters['moduleA/getState']).toBe(2)
     })
   })
--- a/test/unit/modules.spec.js
+++ b/test/unit/modules.spec.js
@@ -1,4 +1,5 @@
-import { nextTick } from 'vue'
+import { h, nextTick } from 'vue'
+import { mount } from 'test/helpers'
 import Vuex from '@/index'
 
 const TEST = 'TEST'
@@ -123,6 +124,49 @@ describe('Modules', () => {
 
       store.commit('a/foo')
       expect(mutationSpy).toHaveBeenCalled()
+    })
+
+    it.only('should keep getters when component gets destroyed', async () => {
+      const store = new Vuex.Store()
+
+      const moduleA = {
+        namespaced: true,
+        state: () => ({ value: 1 }),
+        getters: {
+          getState: (state) => state.value
+        },
+        mutations: {
+          increment: (state) => { state.value++ }
+        }
+      }
+
+      const CompA = {
+        template: `<div />`,
+        created () {
+          this.$store.registerModule('moduleA', moduleA)
+        }
+      }
+
+      const CompB = {
+        template: `<div />`
+      }
+
+      const vm = mount(store, {
+        components: { CompA, CompB },
+        data: () => ({ show: 'a' }),
+        render () {
+          return this.show === 'a' ? h(CompA) : h(CompB)
+        }
+      })
+
+      expect(store.getters['moduleA/getState']).toBe(1)
+
+      vm.show = 'b'
+      await nextTick()
+
+      store.commit('moduleA/increment')
+      console.log(store.state.moduleA.value)
+      expect(store.getters['moduleA/getState']).toBe(2)
     })
   })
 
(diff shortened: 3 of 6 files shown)


testuc22:
25

4refael:
The lack of cache for getters is causing heavy performance issues in some situations.
Will this be fixed with the release of Vue 3.2? seems like the new API is merged.

chuski1212:
Hey! Any news about this? Vue 3.2 has been officially released.

kiaking:
@posva I've added effect scope support on getters creation. Could give it a review and see if it looks good? 🙏 

In Vuex 4, we create computed (getters) object during the `resetStore` function, which will be called on `registerModule` method. So, I think this is the only place we need to care. For example, we don't have `watch` feature in module, hence `watch` will always be called outside of the `setup` hook (on store creation), so I don't think we have to worry about it.

test/unit/modules.spec.js
```diff
@@ -124,6 +125,49 @@ describe('Modules', () => {
...
     })
+
+    it('should keep getters when component gets destroyed', async () => {
+      const store = new Vuex.Store()
+
+      const moduleA = {
+        namespaced: true,
+        state: () => ({ value: 1 }),
+        getters: {
+          getState: (state) => state.value
```
posva, line 136: You could also pass a spy and count the number of times it's called

test/unit/modules.spec.js
```diff
@@ -124,6 +125,49 @@ describe('Modules', () => {
...
+      })
+
+      expect(store.getters['moduleA/getState']).toBe(1)
+
+      vm.show = 'b'
+      await nextTick()
+
+      store.commit('moduleA/increment')
+
+      expect(store.getters['moduleA/getState']).toBe(2)
```
posva, line 169: You can also destroy the `vm` and mount it again

src/store-util.js
```diff
@@ -29,30 +29,42 @@ export function resetStore (store, hot) {
...
+      computedObj[key] = partial(fn, store)
+      computedCache[key] = computed(() => computedObj[key]())
+      Object.defineProperty(store.getters, key, {
+        get: () => computedCache[key].value,
+        enumerable: true // for local getters
+      })
     })
   })
 
   store._state = reactive({
```
posva, line 60: Shouldn't this be inside of the `scope.run` too?

posva (commented):
It looks good although I haven't tried locally. I wasn't so sure about creating a new scope and disposing of the old one, is that for HMR? Isn't it possible to just create one and always use the same?

test/unit/modules.spec.js
kiaking: Yeah I tried it but for some reason I couldn't get destroy to work 🤔  I don't remember why it was long time ago but 😳

src/store-util.js
kiaking: Checking if this is needed or not. Either way we can just wrap it anyway if the confirmation will take time, but just be sure 👀

test/unit/modules.spec.js
kiaking: Great idea! Added! 👍

test/helpers.js
```diff
@@ -4,7 +4,7 @@ import puppeteer from 'puppeteer'
 export function mount (store, component) {
   const el = createElement()
 
-  component.render = () => {}
+  component.render = component.render ? component.render : () => {}
```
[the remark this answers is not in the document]
posva: maybe it just needs `()` wrapping

kiaking:
@posva 
> I wasn't so sure about creating a new scope and disposing of the old one, is that for HMR? Isn't it possible to just create one and always use the same?

Maybe we can? If we re-run the same scope, wouldn't the effects stack up? Since we would be re-creating all getters every time a module is added, I thought we should dispose old effects. Not sure 👀 Let me check on it.

test/helpers.js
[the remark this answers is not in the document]
kiaking: Ah... yeah that sounds right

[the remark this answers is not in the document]
kiaking: Fixed 👍

testuc22:
'

test/unit/modules.spec.js
kiaking: Done!

src/store-util.js
kiaking: I think this is working as is. Since in the test, we increment the state and check if the computed gets updated, and it's passing. So reactive object shouldn't have effect on component lifecycle and it gets destroyed simply by GC.

brianrlewis:
Any progress on this? I'm very eager to upgrade to vue 3 but can't do it until vuex getters are cache properly.

phil294:
I appreciate the work that is being put into this and understand it is not exactly a trivial fix, but could you please add a big red warning at the front page of https://next.vuex.vuejs.org/, discouraging the use of Vuex 4? Getters don't seem to cache anything, at all, not just on component destroy, as far as I have observed. I just spent several work days migrating a few codebases to Vue 3, just to realize it isn't usable yet. In some cases, performance of Vuex went from 100 ms to > 2 minutes, obviously breaking the app.

Not trying to nag, but *please* put an appropriate note at the top of Vuex 4 landing page.

jkosir:
Also appreciate all the work going into this and hopefully this doesn't come off too mean, but how come this is simply getting ignore for 1+ months? The way I see it it's essentially a blocker for Vuex 4.0 adoption in most Vue 3 apps.

tirithen:
I'm part of a team that recently created an app starting from vue 3 and vuex 4. We did not realize these performance issues at first, we assumed that vuex 4 would be a better version of version 3.

Now when our system is built we see that in many cases where we need to use a larger set of input fields, the user interface is locked due to performance issues.

We are currently working on refactoring the code to use as few store bindings that we possibly can. Could you revert the cache changes to the ones from the previous version until that part is more mature?

We like using vuex, we choose it since it is the official store for vue, but if it looks like it takes several months more to resolve this critical problem we might need to switch to redux or similar, so any hints on how long it would take would be greatly appreciated.

I just also wanted to say that we appreciate a lot all the work that has been done on both vue and vuex, for small apps it probably works great, and hopefully, soon it will work for more complex apps as well.

Keep up the good work! :)

kiaking:
@tirithen This is going to be merged soon. Sorry it's taking long but meanwhile, you can build this branch directly and apply the fix manually.

tirithen:
@kiaking That sounds wonderful! Thank you for the fast reply, I'll definitely give it a try to use a build from this branch meanwhile!

Thanks again for all the work being put in!

phil294:
> @phil294 Can you provide example code for this? At least examples in the `examples` folder is working I think. There might be some edge cases?

Sure! I just did: Looks more like a setup problem. I could reproduce it on the examples folder by installing vue *in the example folder*. Here's a bash command to test it out:
`git clone https://github.com/vuejs/vuex && cd vuex && git checkout 4.0 && npm i && pushd examples/classic/counter && npm init -y && npm add vue@next && popd && npm run dev`, then open `http://localhost:8080/classic/counter/`, aaaand: the buttons don't do anything.

I don't know why or what, but I experienced the same issue in other projects of mine when I tried to run it with a locally built 4.0 branch.

blaz-rupnik:
We also tested out this branch because we are in the process of migrating to Vue 3.0 and the reduced performance of getters is a **blocker** for us.

I tested this fix by pointing the vuex package to this branch:
```
"dependencies": {
    "vuex": "git://github.com/vuejs/vuex.git#4.0",
}
```
What happens now is that we have reactivity issues. In one of our components we have:
```
<div class="application__sidebar" v-if="isRouteValid">
    <sidebar />
 </div>
.
.
.
computed: {
    ...mapState([
        'isRouteValid',
    ]),
}
```
So `isRouteValid` is in our store and `false` by default. Once you enter the site we use` router.afterEach` for updating the `isRouteValid` to `true`. But the component doesn't rerender now. I checked if maybe mutation wasn't triggering but it does. Hope this can help as a replication procedure.

phil294:
@kiaking am I right to assume this is still broken? Are there any plans on mending this issue and releasing the fix? Can we help you anyhow? I'm not sure I can figure out this dependency problem in any reasonable amount of time, but I could try. Do you need a PR?
We would love to start using Vue3, finally :-/

jkosir:
No this is still not fixed properly and even this fix isn't released. In Vuejs live UK conf (https://twitter.com/vue_london) on October 20th 2021 every stage-management discussion was focused into Pinia with no mention of any future development of Vuex. 

In Q&A there was a question about this issue specifically, we were told it will be looked into soon™️

thomasdao:
Not sure if anyone has experience with Pinia, does it have the problem with getters like Vuex?

bill-kasa:
Any updates?

nbgraham:
>this fix introduced another bug, breaking reactivity (even worse than before)

#2102 @phil294

Could you be more specific about how this is breaking or provide reproduction code?
I know you provided some steps above,
```
git clone https://github.com/vuejs/vuex && cd vuex && git checkout 4.0 && npm i && pushd examples/classic/counter && npm init -y && npm add vue@next && popd && npm run dev, then open http://localhost:8080/classic/counter/
```
 but that seems like it might be caused by some weird bundling 
>I think this is because Vuex and the Project are using different Vue bundle


I wasn't able to reproduce this issue in a new project. I assumed it meant that changes to getters would not "trigger" updates on `computed`s, but all the reactivity and caching worked fine for me.
I created a new vue app (`npm init vue@latest` which was version 3.2.31) and added vuex from the commit right after the #1883 merge (b2f851f427aa872d1e4f5a4774e07c4c69562789)

```js
const store = createStore({
  state() {
    return {
      count: 0,
    };
  },
  getters: {
    count(state) {
      console.log("computing count getter");
      return state.count;
    },
    countPlusOne(state, getters) {
      console.log("computing count plus one getter");
      return getters.count + 1;
    },
  },
  mutations: {
    increment(state) {
      state.count++;
    },
  },
});
```


```js
<script setup lang="ts">
import { computed } from "vue";
import { useStore } from "vuex";

const store = useStore();
const stateCount = computed(() => store.state.count);
const gettersCount = computed(() => store.getters.count);
const gettersCountPlusOne = computed(() => store.getters.countPlusOne);
const increment = () => store.commit("increment");
</script>

<template>
  <div>
    <div>State Count: {{ stateCount }}</div>
    <div>Getters Count: {{ gettersCount }}</div>
    <div>Getters Count Plus One: {{ gettersCountPlusOne }}</div>
    <button @click="increment">Increment</button>
  </div>
  <template />
</template>
```

phil294:
> Could you be more specific about how this is breaking or provide reproduction code? [...] I assumed it meant that changes to getters would not "trigger" updates on `computed`s

Yes that is exactly what was going on. I also gave up and migrated to Pinia, so I cannot give you up to date code on that anymore, sorry - apart from the reproduction instructions you already saw.

> but that seems like it might be caused by some weird bundling

Yeah well but I guess that's the problem then? Or are you suggesting these are two separate bugs? I don't know, but I assumed otherwise.

rendomnet:
This issue still not fixed?

PR #1892, opened 2020-12-06
docs(zh): add vuex translations for vuex 4

huaxhe:
1. Translated into Chinese based on previous version of the document.
2. Add menu configuration.


Alanscut (changes requested):
It looks like the file docs/zh/api/index.md is still english now.

kadirselcuk (approved):

PR #1896, opened 2020-12-10
Add a generic type for getters too.

mat813:
This allows you to write something like this, for example, in a vuex
module, that lives in a subdirectory of the store, where it is most
interesting:

```ts
import { GetterTree } from 'vuex';

import type { State as RootState, Getters as RootGetters } from '../';
import type { State } from './state';

export type Getters = {
  loggedIn: boolean;
  roles: string[];
  isAdmin: boolean;
};

export const getters: GetterTree<State, RootState, Getters, RootGetters> = {
```

And you get correct type inference in your editor.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -9,11 +9,11 @@ import { createLogger } from "./logger";
 export * from "./helpers";
 export * from "./logger";
 
-export declare class Store<S> {
+export declare class Store<S, G = any> {
   constructor(options: StoreOptions<S>);
 
   readonly state: S;
-  readonly getters: any;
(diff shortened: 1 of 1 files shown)


mat813:
And then, in your actual store that only imports modules, you would do something like:

```ts
import { createStore } from 'vuex';

import * as auth from './auth';

export type State = {
  auth: auth.State;
};

export type Getters = {
  auth: auth.Getters;
};

export const store = createStore<State, Getters>({
  modules: {
    auth,
  },
});
```

(fun fact, with that, the compiler found a typo in some of my code that passed tests anyway.)

PR #1911, opened 2020-12-22
Update .gitignore

evan-lin9:


Kocal:
This kind of exclusion should rather be done in a global `.gitignore` file, see https://docs.github.com/en/free-pro-team@latest/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer.

kiaking:
Yeah I agree with @Kocal. Please add those to the global gitignore 🙌

PR #1916, opened 2020-12-23
docs(typescript-support): fix .js in comment

peter50216:
The extension should be .ts instead of .js, since the file is in TypeScript.


kiaking:
Oops thanks a lot! 🙌

PR #1920, opened 2021-01-02
Add support for preserveState options

### LINKED ISSUE #1675: Dynamic module state not created with hot loading (preserveState set to true)
### Version
3.1.2

### Reproduction link
[https://codesandbox.io/s/compassionate-hermann-msyze](https://codesandbox.io/s/compassionate-hermann-msyze)

### Steps to reproduce
1. Create a module with _vuex-module-decorators_
2. Set the `dynamic` and `preserveState` flags to `true`
    ```ts
    @Module({ dynamic: true, name: 'SomeModule', store, preserveState: true })
    ```
3.  Attempt to access a property of the module, retrieved with `getModule` function to initialise the module from vuex-module-decorators
     ```js
    const someModule = getModule(SomeModule);
    someModule.someProperty;
    ```
4.  Note `someProperty` will be `undefined`

### What is expected?
When installing a dynamic module and there is no state present, a default state should be created

When installing a dynamic module and there is no state present, the module is not registered causing reads to the state for that module to fail.

---
NOTE: Was unable to reproduce the issue cleanly in CodeSandbox so the reproduction is not working, but the key aspects are still there.

Upon investigation, it was discovered the [_vuex-module-decorators_](https://github.com/championswimmer/vuex-module-decorators) lib [passes their `preserveState` flag down](https://github.com/championswimmer/vuex-module-decorators/pull/102) to the vuex layer.  When the dynamic module is installed with the `installModule` function and the `hot` option is enabled, the current implementation skips creating any state for the new module.  This introduces side effects, where the module is registered in the vuex store but the state can't be retrieved if there was no state previously present.

The [_vuex-persist_](https://github.com/championswimmer/vuex-persist) lib is in use, but it is used to hydrate any past state down to _vuex_.  The store, while using the _vuex-persist_ lib, has the data initialised in the following order.

1. The store is created by _vuex_ and any known modules are initialised with their default values.
2. If there are data in the local storage within the browser, it is loaded into the store state replacing the default state data.
3. Dynamic modules are added as they are needed.

Upon modifying the `installModule` function, to check if the state of the new dynamic module exists, the problem is resolved.  Essentially, if the state of the module doesn't currently exist in the `parentState` within the store, then the default module state is created and registered.

One set of changes that fix this case are as follows.

In [`src/store.js`](https://github.com/vuejs/vuex/blob/dev/src/store.js) the following code seems to be the offending code.

```js
// set state
if (!isRoot && !hot) {
  const parentState = getNestedState(rootState, path.slice(0, -1))
  const moduleName = path[path.length - 1]
  store._withCommit(() => {
    if (process.env.NODE_ENV !== 'production') {
      if (moduleName in parentState) {
        console.warn(
          `[vuex] state field "${moduleName}" was overridden by a module with the same name at "${path.join('.')}"`
        )
      }
    }
    Vue.set(parentState, moduleName, module.state)
  })
}
```

When modified, the default state is created when the `moduleName` isn't found in the `parentState`.

```js
// set state
var parentState = getNestedState(rootState, path.slice(0, -1));
var moduleName = path[path.length - 1];
var moduleStateExists = moduleName in parentState
if (!isRoot && (!hot || !moduleStateExists )) {
 store._withCommit(function () {
   {
     if (moduleStateExists) {
       console.warn(
         ("[vuex] state field \""   moduleName   "\" was overridden by a module with the same name at \""   (path.join('.'))   "\"")
       );
     }
   }
   Vue.set(parentState, moduleName, module.state);
 });
}
```

There may be other ways to fix this issue, this is one potential solution that was discovered after some investigation.

filips123:
This adds support for specifying options to `preserveState`, allowing to only skip default module state registration when the state already exists or to merge the existing state with the state from the module. It closes #1675.

### DIFF
--- a/docs/guide/modules.md
+++ b/docs/guide/modules.md
@@ -321,6 +321,10 @@ It may be likely that you want to preserve the previous state when registering a
 
 When you set `preserveState: true`, the module is registered, actions, mutations and getters are added to the store, but the state is not. It's assumed that your store state already contains state for that module and you don't want to overwrite it.
 
+If you want state to be preserved only if the existing state actually exists, you can set the `preserveStateType` to `existing`. In this case, the state from the module will be registered if the existing state does not exist.
+
+You can also set `preserveStateType` to `mergeReplaceArrays` or `mergeConcatArrays`, to merge the existing state with the module state using `deepmerge`, either by replacing arrays or combining them. Individual keys from the existing state will overwrite the default ones from the module.
+
 ## Module Reuse
 
 Sometimes we may need to create multiple instances of a module, for example:
--- a/package.json
+++ b/package.json
@@ -91,5 +91,8 @@
     "webpack": "^4.43.0",
     "webpack-dev-middleware": "^3.7.2",
     "webpack-hot-middleware": "^2.25.0"
+  },
+  "dependencies": {
+    "deepmerge": "^4.2.2"
   }
 }
--- a/src/store.js
+++ b/src/store.js
@@ -2,6 +2,7 @@ import applyMixin from './mixin'
 import devtoolPlugin from './plugins/devtool'
 import ModuleCollection from './module/module-collection'
 import { forEachValue, isObject, isPromise, assert, partial } from './util'
+import deepmerge from 'deepmerge'
 
 let Vue // bind on install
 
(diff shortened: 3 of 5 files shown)


carsoli:
Can we somehow get the attention of the core-reviewers to this issue/PR?

ReWWeR:
When it is expected to be merged?

PR #1930, opened 2021-01-31
Update forms.md

jeremybatesDC:
the two-way computed property solution is SO much simpler that maybe it should be the first suggestion (unless it's really not a best practice)

### DIFF
--- a/docs/guide/forms.md
+++ b/docs/guide/forms.md
@@ -2,6 +2,31 @@
 
 <div class="scrimba"><a href="https://scrimba.com/p/pnyzgAP/cqKRgEC9" target="_blank" rel="noopener noreferrer">Try this lesson on Scrimba</a></div>
 
+## Two-way Computed Property
+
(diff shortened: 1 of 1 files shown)


kiaking:
Thanks for the PR!

I get your point. This is tough decision. I think first having a verbose way describes a bit better understanding on how Vuex works. Which is, you should simply commit mutations on form handling. "Two-way Computed Property" is simpler approach, though it's really a short hand way of doing the exact same thing.

"Two-way Computed Property" is not anti-pattern definitely. However, since the first section (the Vuex way) is not that long, and I think people wouldn't miss the "Two-way Computed Property" coming after, I think keeping it as it is slightly better. Guide users from verbose way to more advanced usage.

I'll leave this issue open for a while to see if can get more feedback on this one.

jeremybatesDC:
thanks @kiaking! Agreed. I've been digging in more this week and absolutely I see what you're saying.

PR #1949, opened 2021-03-18
feat: setup devtools integration

### LINKED ISSUE #1942: Vuex 4 devtools integration
The API for Vue Devtools 6.x is now documented [here](https://devtools.vuejs.org/plugin/api-reference.html).

I open this issue so we can keep track of the Vuex 4 integration.

- [x] Custom inspector with the module tree on the left & module state/getters on the right
- [x] Timeline layer: mutations
- [x] Inspect state after a mutation
- [x] Timeline layer: actions (with event groups)

Akryum:
Closes #1942


posva:
Screen Shot 2021-03-24 at 15 44 58
Screen Shot 2021-03-24 at 15 44 36


Haven't tested nested modules yet

Akryum:
Actions logging
image

Akryum:
Store filtering
image

Akryum:
@kiaking  I've moved the functions from `store.js` to `store-util.js` https://github.com/vuejs/vuex/pull/1949/files#diff-88d13d5b461e04e0a8ffabde604061f24ac756fb0a1f24ce281cf7503c4c5158L253

src/plugins/devtool.js
```diff
@@ -1,26 +1,264 @@
...
+        }
+      })
+    }
+  )
+}
+
+/**
+ * Extracted from tailwind palette
+ */
+const COLOR_PINK_500 = 0xec4899
```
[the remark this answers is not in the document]
kiaking: Oh OK cool! I'll remove it then 👍  I wasn't sure if Devtools was doing some ultra magic to pick these colors up some how haha

package.json
```diff
@@ -54,6 +54,9 @@
   "peerDependencies": {
     "vue": "^3.0.2"
   },
+  "dependencies": {
+    "@vue/devtools-api": "^6.0.0-beta.7"
```
posva, line 58: If it is added as a dependency, shouldn't we mark it as external too? I will need to follow the same on Vue Router. Currently it is not marked as external and therefor is only a dev dependency

kiaking: I thought this was here to support Type imports in TS...? In that case I think it should be the same in Router too. Am I correct @Akryum ?

Akryum:
> Also, what do you think about positioning actions on top of mutations in Timeline view. It's really a personal preference but for me it kinda feels more natural to have actions timeline on top.
> 
> Especially when actions are async, having timeline range bar on top of mutations dots might look better...? Again, just my personal preference though 😳

I don't know, I also thought about it but then the actions layer can grow its height where the mutations will not. So it will probably be pushed far at the bottom if there are a lot of parallel actions going on.

PR #1953, opened 2021-03-22
feat(types): strictly typed

kotamat:
Using StoreObject instead of State allows for more
type-safe calls to the Vuex API from client.

* `store.dispatch()`, `store.commit()` has namespaced-typed params and their payloads
  * the params links to definition on VS Code
* The inside of StoreObject-calls(ex,  commit()) may have
  `any` type
* mapXxx() are not changed.
* Not supported `root`, `namespaced` flags 

4f3d2b20c56f96ba1e23a6b9a77c60e7
34510fb278105f5d3e755a367be76eeb

and changed code formats.

# Please see 
* the naming( because of poor English skill..)
* formats ( default prettier format attached )
* some omissions (don't know well whole Vuex codes)

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "semver": "^7.3.4",
     "start-server-and-test": "^1.12.0",
     "todomvc-app-css": "^2.3.0",
-    "typescript": "^4.1.3",
+    "typescript": "^4.2.2",
     "vitepress": "^0.11.5",
     "vue": "^3.0.5",
     "vue-loader": "^16.1.2",
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -3,31 +3,82 @@ import { App, WatchOptions, InjectionKey } from "vue";
 // augment typings of Vue.js
 import "./vue";
 
-import { mapState, mapMutations, mapGetters, mapActions, createNamespacedHelpers } from "./helpers";
+import {
+  mapState,
+  mapMutations,
+  mapGetters,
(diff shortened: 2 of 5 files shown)


tommie:
Drive-by comment: thanks for working on this! I don't understand the generics involved in `ExtractObjects`, but I'm wondering if it would be possible to skip the separate `S` type param to `useStore` and also extract that from `SO`?

You retracted pull/1943, which seems to be the same thing, but doing the above `S` extraction (and splitting off the reformatting from the individual changes). I'm just curious why you replaced the PR.

kotamat:
@tommie 
`ExtractObjects` transforms `StoreObject` to useful type for args/return-value of `dispatch()` and `commit()`.
like this
```js
const s = {
  modules: {
    foo: {
      modules: {
        hoge: {
          actions: {
            bar: (n: number) => "aaa",
          },
        },
      },
    },
    bar: {
      actions: {
        bal: (x: string) => "aaa",
      },
    },
  },
};
```
to 
```ts
{
  "bar/bal": (x: string) => string,
  "foo/hoge/bar": (n: number) => string
}
```

> I'm wondering if it would be possible to skip the separate S type param to useStore and also extract that from SO?

Yes.



You can skip like this
22e706a7818996ead6cda631dc8788e3
```ts
// store is instance of vuex
  const key: InjectionKey<typeof store> = Symbol('key')
  const usedStore = Vuex.useStore(key)
```

PR #1972, opened 2021-04-21
docs: add TypeScript hint to hot reloading example

caugner:


kiaking (changes requested):
I think we could support both js and ts in the same example. What do you think?

```js
// matches both `.js` and `.ts` file.
const context = require.context("./modules", false, /([a-z_]+)\.(js|ts)$/i)
```

caugner:
Closing this PR, because I have left the company that was using this package and for which I was doing this contribution as part of my daily work. :wave:

Apologies to anybody who may be disappointed, because they counted on this PR getting merged someday. :pray: 

Obviously, if anybody wants to base their contribution on this PR, feel free to do so! No attribution required.

PR #1973, opened 2021-04-21
docs: add info about multiple args for action

estacet:


cuebit:
Thanks for the PR. I think this is lay terms for what already exists. For example, read the following excerpt from the Action docs. Could I therefore ask what your interpretation of this is and whether this just duplicates what's already stated?

> Actions support the same payload format and object-style dispatch

```js
// dispatch with a payload
store.dispatch('incrementAsync', {
  amount: 10
})

// dispatch with an object
store.dispatch({
  type: 'incrementAsync',
  amount: 10
})
```

kiaking:
Yeah, I think this addition is a little too much? But I do understand sometime people gets confused, or not sure how to pass multiple argument. So how about we add a note [at here](https://github.com/vuejs/vuex/pull/1973/files#diff-681b255c8a47359bc04668c27ab9770e5326a66af7931e4e267e8e205cb43aeeR98), stating that a 3rd argument will be ignored and if you want pass multiple arguments, users should pass an object or arrays.

We should do the same for mutations though.

zkriszti:
Explicitly stating that only one argument is accepted as a payload would especially help those who are explicitly looking for information at the Actions section of the docs (that is, using it as a reference, not reading the sections in order). 
It may also help developers who use English as their second language, and it would also make the docs more approachable for beginners.

PR #1984, opened 2021-05-23
perf: optimize _actionSubscribers' hooks execution

peterczg:
As far as I'm concerned, there is no need to filter the array first and then invoke the hooks, just put it together... unless code readability is considered.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -136,8 +136,7 @@ export class Store {
     try {
       this._actionSubscribers
         .slice() // shallow copy to prevent iterator invalidation if subscriber synchronously calls unsubscribe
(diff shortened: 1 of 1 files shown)


cuebit:
Action subscriptions can be either a function or object. In the latter case, since actions are run asynchronously, there needs to be a safety mechanism in place to make sure these hooks exist since they are references, but also as these can be undefined by assignment.

This could be considered a breaking change as it would introduce exceptions where it wouldn't have in some cases.

cuebit:
Thank you for the PR but closing this on the basis of it being a duplicate of #1678 🤝

PR #1986, opened 2021-05-24
fix: no getters displayed on root module + better getters inspector

Akryum:


kiaking (approved):
Wonderful!!! Thanks a lot 👍

PR #1992, opened 2021-06-01
fix(devtools): fix __VUE_PROD_DEVTOOLS__ defined in cjs build (#1991)

### LINKED ISSUE #1991: __VUE_PROD_DEVTOOLS__ is not defined When SSR with prod mode
### Version
4.0.1

### Reproduction link
[https://github.com/cokemine/vuex-reproduce](https://github.com/cokemine/vuex-reproduce)

### Steps to reproduce
clone the repo.
```
yarn
yarn build
yarn start
```

### What is expected?
prod mode works well like dev mode.

4.0.0 works normal

ReferenceError: __VUE_PROD_DEVTOOLS__ is not defined

 at Store.install (E:\vuex-reproduce\node_modules\vuex\dist\vuex.cjs.js:911:50)

CokeMine:
fix #1991
I don't know if I changed it right
But I think __VUE_PROD_DEVTOOLS__ should be replaced to 'false' in the cjs mode.
There should be some small errors here.

### DIFF
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -27,7 +27,7 @@ function createEntries() {
 function createEntry(config) {
   const isGlobalBuild = config.format === 'iife'
   const isBundlerBuild = config.format !== 'iife' && !config.browser
-  const isBundlerESMBuild = /esm-bundler/.test(format)
+  const isBundlerESMBuild = /esm-bundler/.test(config.format)
 
   const c = {
     external: ['vue'],
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -27,7 +27,7 @@ function createEntries() {
 function createEntry(config) {
   const isGlobalBuild = config.format === 'iife'
   const isBundlerBuild = config.format !== 'iife' && !config.browser
-  const isBundlerESMBuild = config.format === 'es'
+  const isBundlerESMBuild = config.format === 'es' && !config.browser
 
   const c = {
     external: ['vue'],
--- a/rollup.config.js
+++ b/rollup.config.js
@@ -27,7 +27,7 @@ function createEntries() {
 function createEntry(config) {
   const isGlobalBuild = config.format === 'iife'
   const isBundlerBuild = config.format !== 'iife' && !config.browser
-  const isBundlerESMBuild = /esm-bundler/.test(config.file)
+  const isBundlerESMBuild = config.format === 'es'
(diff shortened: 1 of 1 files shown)


kiaking:
@CokeMine Thanks for the PR! I've pushed the change. We only need to enable `__VUE_PROD_DEVTOOLS__` in `esm-bundler` build so I've changed it to reflect that.

Just in case, could you try the build and make sure it's fixed at your env?

rollup.config.js
```diff
@@ -27,6 +27,7 @@ function createEntries() {
 function createEntry(config) {
   const isGlobalBuild = config.format === 'iife'
   const isBundlerBuild = config.format !== 'iife' && !config.browser
+  const isBundlerESMBuild = /esm-bundler/.test(format)
```
CokeMine, line 30: I wonder if it should be `const isBundlerESMBuild = config.format === 'es'`
otherwise I cannot build and told me `format is not defined`

kiaking: Oh damn sorry, it should be `config.format` 😓  Fixed!

CokeMine: But why it can pass unit test🤔, and I think isBundlerESMBuild will always be `false`, because no `config.format` can pass the regex

kiaking: Yeah because I think I'm drunk or something... it should be `config.file`... 🤮

kiaking: I think it's finally fixed...!

CokeMine: .......I think `isBundlerESMBuild` is still aways be `false`, it may be `config.file.includes('esm-bundler')`. But I know what you mean, and I think it's better to be `const isBundlerESMBuild = config.format === 'es'`, which is consistent in format

CokeMine:
Yeah, I 've tested in my reproduce repo. I think it is corrent now.

rollup.config.js
kiaking: OK good point. Just updated to not include `esm-browser` builds.

```js
const isBundlerESMBuild = config.format === 'es' && !config.browser
```

CokeMine: Yes! I think it's finished now.

kiaking: 🙌 🙌 🙌 🙌 🙌 

Thank you so much for your patients! 🤝

PR #1994, opened 2021-06-02
fix(dispatch): use promise chaining to fix cancelation in bluebird

### LINKED ISSUE #1993: Bluebird's promise returned by dispatch can't be canceled
### Version
3.6.2

### Reproduction link
[https://codepen.io/davidyuk/pen/GRWQoBY](https://codepen.io/davidyuk/pen/GRWQoBY)

### Steps to reproduce
Open the console and see that "canceled" is not printed after calling `cancel` on dispatch's promise

### What is expected?
To have the promise returned from action be canceled

It doesn't getting canceled

---
Related issues:
https://github.com/vuejs/vuex/issues/1856
https://github.com/vuejs/vuex/issues/1824

It was working fine before, but get broken in https://github.com/vuejs/vuex/commit/6ebbe64c5821d19e55a41dc8b1d81cfce6cbd195

The issue is reproducible after https://github.com/vuejs/vuex/releases/tag/v3.4.0 release.

I have prepared a fix and will publish it soon.

davidyuk:
fixes #1993 

Instead of creating a new promise, I'm proposing to use `Promise.resolve` to cast thenable returned by user's action to `Promise` (according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve)) and then to chain it with action subscribers. Doing this way will allow Bluebird (if it is used as `Promise` polyfill) to store references between promises for proper cancelation.

Also, I think `slice` should be called also for `after` and `error` to protect users from potentially breaking the iteration over subscribers by calling `unsubscribe` in handlers (the same way as it done for `before`).

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -153,15 +153,14 @@ export class Store {
       ? Promise.all(entry.map(handler => handler(payload)))
       : entry[0](payload)
 
-    return new Promise((resolve, reject) => {
-      result.then(res => {
+    return Promise.resolve(result)
+      .then(res => {
         callActionSubscribes('after')
-        resolve(res)
+        return res
       }, error => {
         callActionSubscribes('error', error)
-        reject(error)
+        throw error
       })
-    })
   }
 
   subscribe (fn, options) {
--- a/src/store.js
+++ b/src/store.js
@@ -133,17 +133,21 @@ export class Store {
       return
     }
 
-    try {
-      this._actionSubscribers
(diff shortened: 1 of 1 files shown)


davidyuk:
We are cleaning up our repos, so I would close this PR. I think Bluebird is quite old-fashioned and shouldn't be used in a new project.

PR #2017, opened 2021-06-30
feat: registerModule unregisterModule hasModule support '/' syntax

shunyue1320:


kiaking:
Thanks for the PR! However, we can't do this because module can have `/` in the name (not namespaced, the name it self can contain `/`). So passing in array is required.

Issue #2021, opened 2021-07-12
ERROR: 'Maximum recursive updates exceeded' while use 'vuex' in app.config.errorHandler


LitStronger:
### Version
3.1.4

### Reproduction link
[https://codesandbox.io/s/cranky-snyder-97ll0?file=/src/main.js](https://codesandbox.io/s/cranky-snyder-97ll0?file=/src/main.js)

### Steps to reproduce
open the link, then see the console log.

### What is expected?
app.config.errorHandler()  execute one time

### What is actually happening?
Maximum recursive updates exceeded

LitStronger:
> Are you wanna able to mutate a Vuex state while accessing non-existing nested property of object without recursively called warning?

yes，when the error occurs
, app.config.errorHandler() should execute one time

LinusBorg:
Workaround:

```js
app.config.errorHandler = async (err, vm, info) => {
  console.log(err, vm, info);
  await nextTick()
  store.commit("increment");
};
```

The underlying issue likely related to these two points:

1. the error handler is called during render of `App.vue`(as that's when the error happens), so any reactive state accessed during that time is registered as a render dependency
2. The mutation seems to *mutate* this state, and Vuex might internally *read* the state during that.

There's a couple of things to explore:

1. **Vuex** if happening as I suspect: Is that correct, and necessary
2. **Vue** Should the global errror handler pause Tracking? Local error handlers should not as you might need reactivity to i.e. trigger a re-render for displaying an error message. The global error handler might want/need to do this sometimes as well ...

If so, we might want to better document how to deal with reactivity inside of error handlers, and how the workaround I shared can be used to get around these potential endless loops.

bobby169:
#332

PR #2022, opened 2021-07-13
docs(installation): remove promise polyfill info

manniL:
Because Vue 3 doesn't support IE11 anyway, I thought we can remove the info regarding a `Promise` polyfill in the docs ☺️ 
I've removed the paragraph in all available languages


bobby169:
#2021

kiaking (approved):
Good catch! yeah you're right. Thank you! And thank you for taking care of all the translations as well 🙌

PR #2031, opened 2021-07-25
Add ActionPayload to ActionHandler payload paramameter

maurer2:
Hello, I noticed, that the second parameter of ```Action``` doesn't have the ActionPayload type.

Cheers


weier910:
你的邮件我已收到！

PR #2036, opened 2021-08-09
Update how to install the store as a plugin

kevinjbayer:


docs/guide/index.md
```diff
@@ -34,10 +34,8 @@ const store = createStore({
   }
 })
 
-const app = createApp({ /* your root component */ })
-
 // Install the store instance as a plugin
-app.use(store)
+createApp({ /* your root component */ }).use(store)
```
Alanscut, line 38: creating a Vue instance is not a part of install a plugin, there is no need to bind the creation of Vue instance with the installation of store plugin.

kiaking:
Thanks for the PR! But however, I think we should keep it as is, since we separate call of `use` method in Vue Router guide too.
https://next.router.vuejs.org/guide/#router-view

It might be subjective but I think separating create and use reads better here than making it single line.

kevinjbayer:
No problem, Kia! I should have mentioned that the main reason I created the PR was because I could only get the store to load this way. It wouldn't load when using the current way the docs show. Maybe it's because I'm using typescript or some other setup, I'm not sure. Same with Vue Router: `createApp(App).use(router).use(store, key).mount('#app')`.

kiaking:
Hah, that's odd. I mean

```js
const app = createApp(...)

app.use(...)

// is exact same as doing

createApp(...).use(...)
```

So it must work 👀

PR #2038, opened 2021-08-13
translation bug

Benbinbin:
翻译和英文版不一样


Alanscut (approved):
Thanks! This is the error caused by me 😅

PR #2043, opened 2021-08-23
docs: Change `funciton` to `method`

Yukon123:
`commit` and `dispatch` are two methods of store

 const store = useStore()

 increment: () => store.commit('increment'),


kiaking (approved):
Thanks a lot!

Issue #2045, opened 2021-08-23
Vuex 4 doc Nav-dropdown **language** didn't go to matched page as expect


Yukon123:
### Version
4.0.2

### Reproduction link
[https://next.vuex.vuejs.org/guide/state.html](https://next.vuex.vuejs.org/guide/state.html)

### Steps to reproduce
Hover Nav-dropdown language and click 中文

### What is expected?
Go to https://next.vuex.vuejs.org/zh/guide/state.html

### What is actually happening?
Go to https://next.vuex.vuejs.org/zh/index.html

Alanscut:
updating vitepress to 0.14.0 may fix this bug.

PR #2049, opened 2021-08-31
docs: update vitepress

Alanscut:
Update vitepress to `0.14.0`, closing #2045.  Now changing the language, the sidebar will stay in place instead of redirecting the homepage: https://next.vuex.vuejs.org/zh/index.html.


kiaking:
@Alanscut Sorry I've udpated Vue in other PR and that made this conflict. Could you re open the PR? Also, it would be super cool if you can test out the latest VitePress (it's 0.19 right now 😅

PR #2070, opened 2021-10-13
docs: update vitepress

Alanscut:
Based on #2049, update vitepress to `0.20.0`, closing #2045.


Alanscut:
Hi @kiaking , the reason why netlify deploy failed is that node version is not compatible with vitepress 0.20.0, so

> That's strange, `yarn docs:build` success in my win10 and mac, but failed in the Netlify CI.....

Never mind.... the node version of netlify is set to v12.18.0, which is not compatible with vitepress 0.20, so I upgrade vitepress to 0.19

PR #2081, opened 2021-11-02
feat: add named export in cjs format file

zhangyuang:
`export default` has many bugs when used with `commonjs`, `vuex` use export default object in `vuex.cjs.js` which is not like `vue-router-next`. It cause the below code cannot get correct value in the lastest vite version

```js
import { createStore } from 'vuex'
console.log(createStore === undefined)
```

### DIFF
--- a/src/index.cjs.js
+++ b/src/index.cjs.js
@@ -16,3 +16,16 @@ export default {
   createNamespacedHelpers,
   createLogger
 }
+
+export {
+  Store,
+  storeKey,
+  createStore,
+  useStore,
+  mapState,
+  mapMutations,
+  mapGetters,
+  mapActions,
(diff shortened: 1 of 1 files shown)


zhangyuang:
`vuex.cjs.js` export code

image

zhangyuang:
`vue-router-next` export code
image

zhangyuang:
cc @kiaking , the problem affect users who used the latest vite

sodatea (approved):
FYI, `vue-router` has named exports in v4 cjs build.
So I think it would be more consistent and intuitive if vuex includes them too.

https://unpkg.com/browse/vue-router@4.0.12/dist/vue-router.cjs.js#L3439
/cc @kiaking

PR #2084, opened 2021-11-09
Define state when preserved is true, but there is nothing to preserve

### LINKED ISSUE #2083: Define state when preserved is true, but is nothing to preserve
### Version
3.4.0

### Reproduction link
[codesandbox.io](https://codesandbox.io/embed/optimistic-pine-hqsbc?fontsize=14&hidenavigation=1&theme=dark)

### Steps to reproduce
The fast way to verify this bug is creating a project from scratch, adding vuex, creating a module (I did it in a separate file), setting the namespace to true, registering it like "store.registerModule('bar', module, {preserveState:true})". Now at a vue file call from a a method "this.$store.dispatch('bar/setValue', {name:'foo'})".

### What is expected?
The value is to be set correctly.

Throws a error. 

[Vue warn]: Error in v-on handler: "TypeError: Cannot read properties of undefined (reading 'user')"

---
Vuex handles modules defined in the constructor options different from those registered with store.registerModule. For registerModule calls, the modules pass for a variety of checks before being embedded in the global store.

When checking in "installModule" if the module should be at the root and if the "state" should not be preserved is not observed that if "hot" is set to true, but the state was not defined before throws an error inside the mutation called to change state like "Cannot read properties of undefined (reading 'foo')".

That occurs because it was preserved a no existing state. That's why we should check if the state already exists if so, it will be preserved otherwise we set the default from the module. 

I need the value to be preserved due to the vuex-persistedstate that I'm using and I don't want to get overridden, but at the same time if no state is to be preserved the module default state should be defined.

thiagosantos:
Vuex handles modules defined in the constructor options different from those registered with store.registerModule. For registerModule calls, the modules pass for a variety of checks before being embedded in the global store.

When checking in "installModule" if the module should be at the root and if the "state" should not be preserved is not observed that if "hot" is set to true, but the state was not defined before throws an error inside the mutation called to change state like "Cannot read properties of undefined (reading 'foo')".

That occurs because it was preserved a no existing state. That's why we should check if the state already exists if so, it will be preserved otherwise we set the default from the module. 

The fast way to verify this bug is creating a project from scratch, adding vuex, creating a module (I did it in a separate file), setting the namespace to true, registering it like "store.registerModule('bar', module, {preserveState:true})". Now at a vue file call from a a method "this.$store.dispatch('bar/setValue', {name:'foo'})".

Fixes #2083

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -354,16 +354,14 @@ function installModule (store, rootState, path, module, hot) {
       }
       Vue.set(parentState, moduleName, module.state)
     })
-  }
-  else{
-    //set state when preserved/hot is true, but no state for modules were register before
-      if(! (moduleName in parentState) ){
-        store._withCommit(function () {
-          Vue.set(rootState, moduleName, module.state);
-        });
-      }
+  } else {
+    if( !parentState || !(moduleName in parentState) ){
+      store._withCommit(function () {
+        Vue.set(rootState, moduleName, module.state);
+      });
     }
-
+  }
+  
   const local = module.context = makeLocalContext(store, namespace, path)
 
   module.forEachMutation((mutation, key) => {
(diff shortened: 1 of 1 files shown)


thiagosantos:
@cuebit is this patch going to be applied?

PR #2085, opened 2021-11-10
mapState(["key1", {key2 :'key3'})

zhangenming:
`.map(key => ({ key, val: key }))`
`.reduce((acc, cur)=> {acc.push({ key: cur, val: cur}); return acc}, [])`
I'm not aware of the performance issues Is there a relatively big impact(reduce / map)


zhangenming:
or`mapState(["key1"],  {key2 :'key3'}, ...) `

PR #2087, opened 2021-11-17
Updated Installation docs for Vue CLI Installation

deotte:
The current docs don't include information about installing Vuex through the Vue CLI. I couldn't find information on the Vue CLI docs about Vuex, so I figured I'd add it here. I also copied the installation text from the [Vue router docs](https://router.vuejs.org/installation.html#direct-download-cdn) to keep it consistent.

### DIFF
--- a/docs/installation.md
+++ b/docs/installation.md
@@ -21,6 +21,14 @@ Include `vuex` after Vue and it will install itself automatically:
 npm install vuex@next --save
 ```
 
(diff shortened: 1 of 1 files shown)


Alanscut (approved):

deotte:
I'm assuming this isn't necessary anymore because of Pinia becoming the future standard - going to go ahead and close

PR #2096, opened 2021-12-09
Update state.md

jasba24:
Fix typo in the doc


cuebit:
Thanks for the PR but I’m afraid this is not a typo.

jasba24:
Oh yes, I see that it is not a typo, but I think it would be better to put a point in the letters to be more specific and help people who know little English

jasba24:
I mean something like that **i.e**

cuebit:
@jasba24 yes, I think that does add clarity. If you push that change to this PR, I'm happy to re-open it?

PR #2143, opened 2022-03-09
change types for overloads (fix #2142)

### LINKED ISSUE #2142: Changes to typescript overloads in Store instance methods.
### Version
3.6.2

### Reproduction link
[codesandbox.io/s/umtsos](https://codesandbox.io/s/umtsos?file=/src/store/index.ts)

### Steps to reproduce
1. Look up the Union type of the 'path' argument (initModule method)

2. An error appears about the incompatibility of the types of arguments for the registerModule method, the typing of this method does not support the use of the Union type:` string | string[]`

3. Familiarize yourself with the typing of this method by reference and match the union type to the two method declarations:
https://github.com/vuejs/vuex/blob/3.x/types/index.d.ts

For more information, use the resource:
https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#use-union-types

### What is expected?
Support Union type: string | string[] given the already described type overloads of the registerModule method, the following errors will be eliminated when resolving the conflict:

```
1. Argument of type "string | string[]" is not assignable to parameter of type "string[]".
2. Argument of type "string | string[] is not assignable to parameter of type "string".
```

When comparing the type "string | string[]", an error about type incompatibility is thrown, but given the type overloads, it should be possible to put an argument with the union type.

Surof1n:
PL created for resolve #2142

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -24,14 +24,11 @@ export declare class Store<S> {
   subscribeAction<P extends ActionPayload>(fn: SubscribeActionOptions<P, S>, options?: SubscribeOptions): () => void;
   watch<T>(getter: (state: S, getters: any) => T, cb: (value: T, oldValue: T) => void, options?: WatchOptions): () => void;
 
-  registerModule<T>(path: string, module: Module<T, S>, options?: ModuleOptions): void;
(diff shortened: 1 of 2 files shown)


Surof1n:
This problem also exists in Vuex 4 :(

PR #2149, opened 2022-03-25
Make vuex usable in react, react-native

visitsb:
Using Vuex outside of Vue projects looks possible with a few simple changes. Vuex works independently only using few imports from `@vue/reactivity` and `@vue/runtime-core`. Switching these to pure imports removes the need to import them through Vue. 

Additionally, component binding helpers `mapXXX`, `createNamespacedHelpers` only presume `this.$store` is accessible. Switching `this` to `globalThis` from `@vue/shared` allows `this` to be flexibly determined by runtime Javascript's [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis).

@yyx990803, @ktsn, @kiaking - I have managed to use Vuex store successfully in a react-native project after making these changes. The best part is it works according to Vuex documentation. 

It will great if these changes make it to mainline branch. Love to use Vuex over other state management libraries in React ecosystem.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 *.log
 .DS_Store
 node_modules
+.idea
--- a/package.json
+++ b/package.json
@@ -56,7 +56,10 @@
     "vue": "^3.2.0"
   },
   "dependencies": {
-    "@vue/devtools-api": "^6.0.0-beta.11"
+    "@vue/devtools-api": "^6.0.0-beta.11",
+    "@vue/reactivity": "^3.2.31",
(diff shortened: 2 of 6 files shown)


visitsb:
@yyx990803 , @ktsn , @kiaking I understand Vuex is now in maintenance, and recommending Pinia. 

Since this PR doesn't look like going to merge into Vuex anytime soon, I have gone ahead and publish a fork under [@visitsb/vuex](https://github.com/visitsb/vuex) so that non-Vue.js applications such as React, ReactJS and React Native can use Vuex.

PR #2156, opened 2022-04-07
style: delete unused variable of shopping-cart

lxchuan12:
import { currency } from './currency'

unused


kiaking:
Oops, thanks!

Issue #2181, opened 2022-06-17
VueX store loses reactivity if shared using window on page with multiple Vue bundles


stam:
### Version
3.4.0

### Reproduction link
https://github.com/stam/multiple-vue-apps

### Steps to reproduce
Clone the mentioned repo
npm i
npm run serve
open ./index.html
Click on the count++ buttons

### What is expected?
I expect both counters to update and reactivity to be intact

### What is actually happening?
Only one counter updates, as the Vuex store in the other app is not reactive

---
We have 2 vue bundles on the same page. They share their VueX store using window.globalStore, but when you press count++ only one app is updated.

The obvious fix would be to just use 1 vue bundle. However this doesn't work in my company's usecase: We are gradually migrating from a legacy server side rendered application to a SPA. Small parts of the application are being converted to Vue, where the legacy application loads the specific Vue bundles on demand. However multiple Vue bundles can live on the same page, and we want to share some state as we don't want to duplicate API calls.

TimvdEijnden:
I'm facing the same issue. We can create a new store and call 

```js
newStore.registerModule('users', { 
  ...existingStore._modules.root._children.users._rawModule, 
  state: existingStore.state.users}
);

Then the reactivity works. But it's still not ideal and causes issues when using Vuex in strict mode.

Did you find a solution @stam ?

@yyx990803 @kiaking Would there be an option to manually add the reactivity to the 2nd vue instance?

**Comment by stam** [comment]
@TimvdEijnden I haven't found a fix, we're now just not sharing the store and instead are refactoring to an SPA.

The core issue is probably because reactivity is managed by the Vue singleton, and the Vue instance in bundle B does not keep track of the VueX store in bundle A

**Comment by rodhoward** [comment]
I'm not sure if what we're experiencing is exactly the same issue but we are trying to lazy load routes (bundles) and then using 
```
 if (store.hasModule(key)) return;
 store.registerModule(key, componentStateObject);
```
To register the extra route components when the page/route loads.

Its working great except that we seem to lose reactivity on our getters after we register the first dynamic module.

Versions:
"vue": "^2.6.14",
"vuex": "^3.6.2",
"vue-router": "^3.5.3",
"@vue/cli-plugin-babel": "^3.12.1",
"@vue/cli-plugin-eslint": "^3.12.1",
"@vue/cli-plugin-unit-jest": "^4.5.17",
"@vue/cli-service": "^4.5.17",
"@vue/compiler-sfc": "^3.2.20",
[...]

Issue #2197, opened 2022-11-15
Getters - Reactivity broken after registering new module


lordolive:
### Version
Vue:  3.2.45
Vuex: 4.1.0

### Reproduction link 
https://github.com/lordolive/vuex4-getters-reactivity

### Steps to reproduce
* Clone the repo
* npm i
* npm run dev

Then in website :

* Click the first button, the getter is reactive and the count increments.
* Click the second button to register the module
* Click the first button, the getter is no longer reactive and the count does not change.

### What is expected?

The getter reactivity should still work and the count in the button should increment.

### What is actually happening?
The getter accessed before the call to this.$store.registerModule() is no longer reactive, but the one that was not accessed before is still reactive

![image](https://user-images.githubusercontent.com/30758833/201994278-1309a2b5-1b83-4953-ac4d-14fa478f8012.png)

### Additionnal comments
I first tried with MapGetters and had the same problem.
Using $store.getters.count directly in the template solve the issue, but in my real application with multiple modules it is simply not pratical, especially considering that we are migrating a lot of code from Vue 2 which was working with mapGetters.

jkosir:
Experiencing the same issue, the problem is the replacement and stopping of `store._scope` in [resetStoreState](https://github.com/vuejs/vuex/blob/b7f471aef9265555f4094b313e241b4060812cfb/src/store-util.js#L30) which is called when registering a new module. 

This stops currently active effectScope (`store._scope`) onto which all getters are bound and registers them again, which will cause any effect (computed, watch, e.g. `count1` in reproduction example linked above) to become inactive and simply stop reacting to state changes.

A workaround (non-feasible, but still) is to subscribe to `onScopeDispose` of `store._scope` and re-register computed/watch effects you need, e.g. like this:
```
store._scope.run(() => {
	installWatcher()
})

function installWatcher () {
	onScopeDispose(() => {
	    installWatcher()
	})
	watch(() => ...)
})
```

@kiaking @posva  (since you originally added and reviewed this in https://github.com/vuejs/vuex/pull/1883) unless I'm missing something the replacement of `store._scope` shouldn't be needed ([here](https://github.com/vuejs/vuex/blob/0501bcad2aaac2479e828a6aa512b2f79c8ddc7f/src/store-util.js#L83))? If this is done then any effect referencing getters will become inactive and afaik there's no way to reactivate it from store.

catalin-bratu:
Any ETA till this bug will get fixed?

[...]

PR #2201, opened 2022-12-09
fix: keep computed getter be reactive after registering new modules

Azurewarth0920:
close: #2181 #2197 

The computed getters will be unreactive after registering modules,

Approach:

When stopping effectScope, only stop the effect that has no dependencies.


test/unit/modules.spec.js
```diff
@@ -925,4 +925,31 @@ describe('Modules', () => {
       /getters should be function but "getters\.test" in module "foo\.bar" is true/
     )
   })
+
+  it('module: computed getter should be reactive after module registration', () => {
```
Azurewarth0920, line 929: The test will fail when removing this line: https://github.com/vuejs/vuex/pull/2201/files#diff-d7ab881f3cc9bbff70c15e5574df58176996a0b834bc174d3f3cea7a64ad8a95R87 .

Azurewarth0920:
Hi! @kiaking @posva 

I think this PR may fix the broken feature in v4.1.0.
if you have time, can you review this PR?

Azurewarth0920:
@1085383233 
https://github.com/vuejs/vuex/pull/2201#issuecomment-1350325599
Hi! Can you provide a test case about the problem?

Azurewarth0920:
@1085383233 

> Hi! There may be a problem with the unregisterModule function.

Is the `ReactiveEffect` with the module kept alive even if the module is unregistered?

melody001du:
> 

Yeah, but I didn't test it.

melody001du:
> @1085383233
> 
> > Hi! There may be a problem with the unregisterModule function.
> 
> Is the `ReactiveEffect` with the module kept alive even if the module is unregistered?

@Azurewarth0920 Yeah, but I didn't test it

melody001du:
> @1085383233
> 
> > Hi! There may be a problem with the unregisterModule function.
> 
> Is the `ReactiveEffect` with the module kept alive even if the module is unregistered?

@Azurewarth0920 Yeah, but I didn't test it.

catalin-bratu:
Any news on this?

The bug is really a bummer.

Azurewarth0920:
@catalin-bratu 
This PR is waiting for reviews.

At this time, the workaround for my project is that downgrade the vuex to v4.0.2.

enkil2003:
@Azurewarth0920 is there any known bug for this fix you want to merge?
any updates on this?

alecgibson:
I've raised an alternative solution in https://github.com/vuejs/vuex/pull/2234

alecgibson:
I've closed my alternative solution, which fails this test case:

```js
  it('module: computed getter should be reactive after module unregistration', () => {
      state: {
        foo: 0
      },
      getters: {
        getFoo: state => state.foo
      },
      mutations: {
        incrementFoo: state => state.foo++
      }
    })

    const computedFoo = computed(() => store.getters.getFoo)
    store.commit('incrementFoo')
    expect(computedFoo.value).toBe(1)

    store.registerModule('bar', {
      state: {
        bar: 0
      }
    })
    store.unregisterModule('bar')

    store.commit('incrementFoo')
    expect(computedFoo.value).toBe(2)
  })
```

@Azurewarth0920 do you think it's worth adding this test case to your PR? (Your code passes it 👏🏼 )

src/store-util.js
```diff
@@ -45,6 +47,10 @@ export function resetStoreState (store, state, hot) {
 
   scope.run(() => {
     forEachValue(wrappedGetters, (fn, key) => {
```
suzhenskk, line 49: trytry

weier910:
你的邮件我已收到！

PR #2204, opened 2022-12-26
docs(zh): update zh docs tip from en docs

fengnzl:
I saw "Pinia is now the new default" in en doc, but not found in zh-cn doc, so I translated it.


13RTK:
Why did the PR not merge now?

PR #2231, opened 2023-08-09
fix : link pointing errors

heygsc:
The original link was not found and the computed attribute was not found in the Japanese version, so it was changed to the English version to avoid errors in the original link.

issue :  https://github.com/vuejs/vuex/issues/2232


weier910:
你的邮件我已收到！

PR #2234, opened 2023-08-31
fix: avoid resetting store state when registering a dynamic module

alecgibson:
Fixes https://github.com/vuejs/vuex/issues/2197

At the moment, when registering a dynamic module, we call `resetStoreState()` just to register the getters for the new module.

It seems unnecessary to reset the entire store state in this case, and this actually also leads to [other issues][1].

This change is based on the test case added in https://github.com/vuejs/vuex/pull/2201

The approach taken in this change is to refactor the getter registration into its own function, and call that new method when registering a dynamic module instead of resetting the store state.

[1]: https://github.com/vuejs/vuex/issues/2197


alecgibson:
Okay I thought this would be cleaner than https://github.com/vuejs/vuex/pull/2201 but it unfortunately doesn't handle module unregistration (which [also clobbers the state](https://github.com/vuejs/vuex/blob/7b024d8bc7ed8e273b8bd853b1e23ec78ab1b045/src/store.js#L247)). I think the approach in this PR would need basically a full rewrite of the `unregisterModule()` method to avoid completely resetting the rest of the store state, which is a pretty big change. Happy to do it if it's a direction that people want to go in, but otherwise let's go for https://github.com/vuejs/vuex/pull/2201

PR #2242, opened 2024-02-19
Add generic type to allow typing getters

apperside:
Added a new generic type to allowing typing the getters


weier910:
你的邮件我已收到！

PR #2251, opened 2024-07-15
Vuex Devtools plugins registered on server which increases memory usage

### LINKED ISSUE #2250: Vuex Devtools plugins registered on server which increases memory usage
### Version

4.0.2

### Describe the bug

Recently upgraded Vuex to 4.0.2 for a Vue 3 migration that uses a custom SSR solution using Express and Webpack.  When running in development mode, the server's memory continued to rise with usage, whereas in production mode, it didn't happen.

Ran a profiler via Chrome's devtools and noticed that in development mode, there were large memory allocations for the __VUE_DEVTOOLS_PLUGINS__.  We also use vue-i18n and vue-router which also register devtools plugins when running in development mode.

However, those other libraries make a check that they are executing in the browser or not in a Node environment, but Vuex does not.  As long as you are running in development mode (even on the server) it registers the devtools plugins.  I believe this is the cause of the sustained memory usage (possibly a leak?).

I will submit a PR to fix this issue.

### Reproduction

1. Use Vuex 4.0.2 with Vue 3 with SSR on a Node server process while running with NODE_ENV=development
2. Make sure the usage includes registering store modules at runtime
3. Notice that as the server processes more requests that use Vuex, the memory used by the server will rise and not come down.
4. Notice that the same thing doesn't happen if you run the server with NODE_ENV=production

### Expected behavior

Vuex not to add devtools on the server (even if NODE_ENV=development), client-side is fine.

### Additional context

_No response_

### Validations

- [X] Follow our [Code of Conduct](https://vuejs.org/about/coc.html)
- [X] Read the [docs](https://vuex.vuejs.org/).
- [X] Check that there isn't already an issue that reports the same bug to avoid creating a duplicate.

urbnjamesmi1:
Fixes #2250

- Add check that code is running in a browser before adding devtools

Would need this fix for 4.0.2 since I cannot upgrade to 4.1.0 due to bugs with deregistering modules at runtime.

### DIFF
--- a/src/store.js
+++ b/src/store.js
@@ -12,6 +12,8 @@ import {
   unifyObjectStyle
 } from './store-util'
 
+const isBrowser = typeof document !== 'undefined'
+
 export function createStore (options) {
   return new Store(options)
 }
@@ -73,9 +75,9 @@ export class Store {
     app.provide(injectKey || storeKey, this)
     app.config.globalProperties.$store = this
 
-    const useDevtools = this._devtools !== undefined
+    const useDevtools = (this._devtools !== undefined
       ? this._devtools
-      : __DEV__ || __VUE_PROD_DEVTOOLS__
+      : __DEV__ || __VUE_PROD_DEVTOOLS__) && isBrowser
 
     if (useDevtools) {
       addDevtools(app, this)


urbnjamesmi1:
@kiaking @yyx990803 Would it be possible to get this merged?

PR #2257, opened 2024-09-25
docs: augment `vue` rather than `@vue/runtime-core`

danielroe:
https://github.com/nuxt/nuxt/issues/28373#issuecomment-2373531197

This updates the docs to use the currently recommended way to augment component custom properties in Vue 3.


weier910:
你的邮件我已收到！

cuebit:
Thanks @danielroe!

PR #2258, opened 2024-10-15
fix: augment vue rather than @vue/runtime-core

joaovitoras:
Follow new way to augmenting global properties

https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties


weier910:
你的邮件我已收到！

PR #2270, opened 2025-12-08
Add types to exports in package.json

yangdan8:


mat813:
With typescript 6.0, it would be great to have this in a vuex 4.1.1.

PR #2275, opened 2026-07-24
fix: AddExports

EndlessKeep:


EndlessKeep:
Without this condition, TypeScript projects using moduleResolution: "bundler" resolve vuex to dist/vuex.mjs instead of the existing type declarations, causing Vuex APIs such as mapState to lose their types and producing downstream inference errors in Vue components.

This change preserves the existing runtime exports while allowing modern TypeScript module resolution to correctly resolve types/index.d.ts.

